PR #13, opened 2026-08-14
Add i18n/localization support (23 Indian languages)

Lihini0202:
### **Purpose**
All 23 language bundles were compiled directly into the JavaScript build. Two consequences followed from that:

Every reader downloaded every language, whether they used it or not - the full translation set shipped in the bundle regardless of which language a session actually rendered in.
Correcting a single word required a full rebuild and redeploy. Because the strings lived in .ts files compiled by Vite, there was no way to fix a wrong translation on a running deployment without going through the build pipeline again.

### **Goals**

- A reader's browser fetches only the language it is actually rendering in, not all 23.
- Correcting the wording of an existing key is a file edit on the deployed server, not a rebuild.
- Adding a language to a running deployment (new folder + a meta.json entry) needs no rebuild either.
- Adding a new key remains a code change either way - no product avoids this, and this PR does not attempt to.

### **Approach**
Translations move from compiled .ts modules to static JSON served from public/i18n/, fetched at runtime via i18next-http-backend.

- src/i18n/resources/<lang>/{common,catalog}.ts (46 files) → public/i18n/<lang>/{common,catalog}.json (46 files). Every existing translation was preserved key-for-key across the conversion; nothing was re-translated or dropped.
- i18n.ts: switched from the eager import.meta.glob (which is what compiled every language into the bundle) to i18next-http-backend, with loadPath resolving i18n/{{lng}}/{{ns}}.json under the app's base path.
- languages.ts: the language list is no longer only a compiled constant. BUNDLED_LANGUAGES remains as a fallback; loadLanguageCatalogue() fetches public/i18n/meta.json at startup and that becomes the list actually offered, so a language can be added to a live deployment without a rebuild. A failed fetch falls back to the bundled list rather than leaving the switcher empty.
- main.tsx: sequenced so the catalogue loads, then i18next initializes against it, then the app renders - awaited, so there's no frame of raw translation keys before the first paint.
- scripts/sync-catalog-i18n.mjs: the administrator-catalog sync script now reads and writes catalog.json instead of generating a .ts module.
- scripts/build-i18n-meta.mjs (new): generates public/i18n/meta.json from the bundled language list, wired into npm run build, with a --check mode so CI can catch drift between the two.
- vitest.setup.ts: preloads the English bundle directly and registers every other language as an empty bundle, since nothing answers an HTTP fetch under jsdom. Without this, any test asserting on rendered text - or switching language - would either see raw keys or hang waiting on a request that never resolves.
- Two of the portal's own security/quality gates required small follow-on fixes: font-src needed data: added (Oxygen UI ships its typefaces as data: URIs), and one explanatory comment was reworded because it contained the literal string the production source-sink scanner checks for.
- No component-level code changes were needed beyond this - every t('key') call site is unaffected, since the key namespace and lookup behaviour are identical to before.

### **User stories**
As a reader of the portal in any of the 23 supported languages, the app only downloads the language I'm using, not all 23.
As an operator, I can correct a wrong translation, or add a new language, on a running deployment without rebuilding or redeploying the application.

### **Release note**
Portal translations are now served as static JSON at runtime instead of being compiled into the application bundle, reducing per-language download size and allowing wording corrections and new languages to be deployed without a rebuild.

### **Automation tests**

**Unit tests:** vitest run - 171/171 passing across 27 files, unchanged from before this PR. I18nKeys.test.ts now reads the shipped public/i18n/en/common.json directly, so it validates the exact file the server serves rather than a compiled intermediate.
**Integration tests:** None added; this is a build-pipeline change with no new runtime surface to integration-test. The existing component test suite continues to pass unmodified against the new loading path via the vitest.setup.ts preload described above.

**Test environment**
Node.js, npm; vitest + jsdom for the unit suite; vite build for the production bundle.

### **Learning**
Looked at how a couple of existing products split "translations compiled into the bundle" from "translations fetched at runtime" - the pattern is consistent across implementations: build-time compilation only for a fallback/default set, with the live catalog resolved from a separate static or API-served resource at startup. This PR applies that same split using i18next-http-backend, the standard i18next runtime-loading plugin for exactly this case.


hasithakn:
@coderabbitai can you create issue for

One thing worth flagging separately, not part of the build failure you reported: npx vitest run currently has 36 failing tests (translated strings coming back as raw i18n keys, e.g. layout.userMenu.signOut instead of "Sign out") — looks like the test setup wasn't updated for the new async i18next-http-backend loading model.

PR #43, opened 2026-08-28
Let users delete their own account from the consent portal

anjuchamantha:
## What this adds

A signed-in portal user can now delete their own account. **Delete my account** sits in the profile menu next to Sign out; confirming it calls `DELETE /scim2/Me`, clears the browser session and lands the user on a public confirmation page.

## Why it needed a new scope

Identity Server protects `DELETE /scim2/Me` with `internal_user_mgt_delete` by default. That same scope also authorizes `DELETE /scim2/Users/{id}` — deleting *other people*. Handing it to portal users so they could delete themselves would let any one of them delete anybody.

So the accelerator's `deployment.toml` overrides that one endpoint to require a much narrower scope instead:

```toml
[[resource.access_control]]
context = "(.*)/scim2/Me"
allowed_auth_handlers = ["OAuthAuthentication"]
secure = "true"
http_method = "DELETE"
scopes = ["account:self:delete"]
```

It is deliberately OAuth-only: self-deletion should come from a user-context token, not Basic auth. `internal_user_mgt_delete` is never granted to portal users, so `/scim2/Users/{id}` keeps its administrator-only protection.

Tenant provisioning registers `account:self:delete` (via a virtual API resource that exists only to mint the scope — it never serves traffic), authorizes the portal application for it, and grants it through the **`dpdp-consent-user` role only**. Administrators do not get it, so the server answers their `DELETE /scim2/Me` with a 403 whether it comes from the portal, curl, or anywhere else. The portal hiding the menu item for them is a convenience on top of that, not the control itself.

## Approval workflows

If an approval workflow is associated with the Delete User operation, the account is **not** deleted when the user confirms. The server records a request and answers `202`, and the account stays fully usable — the user can keep working and sign in again — until an approver acts.

The portal has no way of knowing in advance whether a workflow is configured, so it branches on the response:

| Response | What the user sees |
|---|---|
| `204` | Session cleared, account-deleted page. |
| `202` | Stays signed in. "Your deletion request has been submitted for approval. Your account stays active until an administrator approves it." |
| `400` | "Your account already has a deletion request awaiting approval." The server refuses a second request while one is pending. |

The `400` is keyed off the status code rather than matching the server's message text, since a 400 from this call has no other cause.

Approvers act on the request in **My Account** (`/myaccount`). The Console's Workflow Requests page is a monitoring view — it lists requests and can abort one, but offers no approve action.

## A correction to the role documentation

While testing this, it turned out `dpdp-consent-user` was an **empty role that granted nothing**. Verified against a running server: a user with no portal role at all can sign in, reach the dashboard and manage their own consents on `internal_login` alone. The configuration guide's claim that "every user of the portal needs one of these two roles" was simply wrong.

That matters now the role carries `account:self:delete`. **Users who do not hold it will not see the option**, and since the role previously granted nothing, existing accounts are unlikely to have it. The guide now says so plainly and describes both roles as granting what is *beyond* a plain login, rather than being required for one.

There is also a caveat worth knowing: the two roles' permissions add up, so a user holding **both** `dpdp-consent-admin` and `dpdp-consent-user` does receive the scope and can self-delete. Operators who care should keep administrators out of the user role.

## Migrating existing tenants

Role provisioning now reconciles rather than skipping roles that already exist: it adds whichever required permissions a role is missing, and never removes any, so anything an operator granted by hand survives. That is also the migration path — re-running provisioning (the documented "update the tenant" recovery step) backfills the new scope onto tenants created before this existed. No new tooling, no restart, no role deletion.

## Known limitation

**Deleting an account does not clean up the user's DPDP data.** Their consent records and event subscriptions stay behind, referencing a user who no longer exists. This is called out in the configuration guide as an operator task; cleaning it up is separate work.

Two things were also left out deliberately:

- **Disabling the account while a deletion request is pending.** That needs the `accountDisabled` claim, is reversible only by an administrator, and is a policy decision rather than a bug fix.
- **Showing "deletion requested" in the menu.** There is no self-service API to ask whether the signed-in user has a request pending — the workflow-request APIs are administrative — so the `400` message covers that case instead.

## What changed

**Identity Server side**
- `deployment.toml`: the `[[resource.access_control]]` override above.
- `DPDPApiResourceProvisioningUtil`: registers the account self-service API resource and authorizes the portal application for it.
- `DPDPIdentityExtensionTenantMgtListener`: `account:self:delete` joins the user-role scope list.
- `DPDPConsentPortalRoleProvisioningUtil`: reconciles missing permissions onto existing roles.

**Portal**
- New `features/account` module: the confirmation dialog, the API call, the delete mutation and the public account-deleted page.
- `UserProfileMenu`: the scope-gated menu item, the pending-approval notice and the error messages.
- `App.tsx`: the account-deleted route lives outside the authentication gate, since someone whose account is gone cannot complete a sign-in.
- `apiClient`: a helper exposing the response status, so success and "accepted for approval" can be told apart.
- New i18n strings in English, copied across the 22 fallback locales.

**Tests**
- Unit tests for the status-to-outcome mapping, the pending-approval path keeping the session alive, and the already-requested message.
- Backend tests for idempotent registration and authorization of the new API resource.
- `tests/06-account`: end-to-end deletion against a real server. Destructive and irreversible, so each test creates and signs in as its own throwaway user rather than a shared persona, and removes it afterwards — including approving the pending task its own cleanup raises. The main test accepts either `202` or `204` and asserts the matching UI, so it is correct with or without a workflow configured.
- `tests/04-authorization`: a user is offered account deletion, a consent admin is not.

## Verification

- Full Maven reactor build: **BUILD SUCCESS**, 406 backend tests passing.
- Frontend: typecheck clean, 215 unit tests passing, production build green including the CSP and 23-language i18n verifiers.
- Integration suite typechecks clean and all specs load.
- Exercised against a live WSO2 IS 7.3.0 with an approval workflow configured: the `202` path keeps the user signed in and reports the pending request, the account survives, a second attempt reports the existing request, and a user token is refused when it tries to delete a different account.


hasithakn (approved):

anjuchamantha:
@coderabbitai create an issue to track post user-delete actions. And check the concern of IS deleting consents of a deleted user.

PR #49, opened 2026-08-28
Add completion and polling endpoints with JWS-signed poll deliveries

ThiwankaChanditha:
# Add completion and polling endpoints with JWS-signed poll deliveries

## Purpose

This PR introduces short polling and delivery-completion capabilities to the Event Notification Framework.

Poll deliveries are returned as tenant-aware RS256-signed JWS tokens. Completion reports are authenticated separately using the subscription-level HMAC `event-signature`; completion requests are not JWS payloads.

### 1. Event polling endpoint

Added:

```http
POST /api/dpdp/event-notifications/v1/events/poll
```

The endpoint:

- Resolves the organization from the tenant URL context.
- Identifies the poll subscription using `subscription-id`.
- Restricts polling to the matching organization, group and active poll subscription.
- Supports short polling through `maxEvents` and `returnImmediately`.
- Does not support long polling.
- Returns `moreAvailable` when additional deliveries remain.

### 2. Poll acknowledgement and error handling

Poll requests can update previously returned deliveries using delivery IDs:

```json
{
  "ack": [
    "delivery-id-1"
  ],
  "setErrs": {
    "delivery-id-2": {
      "err": "processing_failed",
      "description": "The event could not be processed"
    }
  },
  "maxEvents": 20,
  "returnImmediately": true
}
```

Acknowledgement and error updates:

- Are scoped to the selected organization, group and subscription.
- Only update deliveries currently in the `pending` state.
- Reject delivery IDs appearing in both `ack` and `setErrs`.
- Are committed within a single database transaction.

### 3. JWS-signed poll deliveries

Every returned poll delivery is represented by its delivery ID and compact JWS:

```json
{
  "moreAvailable": false,
  "sets": {
    "delivery-id": "<base64url-header>.<base64url-claims>.<base64url-signature>"
  }
}
```

The polling implementation reuses the tenant-aware signing mechanism introduced for webhook delivery:

```
Pending poll delivery
    ↓
Load event and subscription context
    ↓
Build complete event envelope
    ↓
Resolve tenant OAuth signing key
    ↓
Sign the envelope using RS256
    ↓
Return compact JWS under the delivery ID
```

The signed payload includes the delivery, event, subscription, organization, group, topic and original event payload.

### 4. Completion endpoint

Added:

```http
POST /api/dpdp/event-notifications/v1/deliveries/{deliveryId}/completion
```

The endpoint accepts a completion report after a consumer processes a webhook delivery:

```json
{
  "completionStatus": "completed",
  "completionEvidence": "https://processor.example/evidence/receipt.pdf",
  "completedAt": 1752624120000
}
```

The implementation:

- Resolves the organization from the tenant URL context.
- Confirms that the delivery belongs to the specified organization and group.
- Verifies the `event-signature` as HMAC-SHA256 over the exact request body.
- Uses the subscription shared secret for verification.
- Stores the completion status, evidence and completion time.

The completion body itself is not JWS-signed.

### 5. Polling configuration

Added configuration for:

```toml
[dpdp_accelerator.event_notifications.polling]
default_return_immediately = true
default_max_events = 20
max_events_limit = 100
request_hmac_validation_enabled = false
```

These settings are exposed through the standard configuration chain:

```
deployment.toml
    ↓
dpdp-accelerator.xml.j2
    ↓
DPDPConfigParser
    ↓
DPDPConfigurationService
    ↓
Event polling service
```

### 6. Database support

Extended poll-delivery persistence to support:

- Delivery-level acknowledgements.
- Error status and error-code persistence.
- Subscription-scoped pending-delivery retrieval.
- Guarded status transitions from `pending`.
- Consistent behavior across H2, MySQL, PostgreSQL and SQLite.

## Security model

| Operation | Security mechanism |
|---|---|
| Poll request | OAuth scope and optional request-body HMAC validation |
| Poll response | Tenant-aware RS256 JWS |
| Completion request | OAuth scope and mandatory request-body HMAC validation |
| Completion body | Plain JSON authenticated by HMAC; not JWS |

## Testing

Tests cover:

- Poll subscription, organization and group validation.
- Poll request HMAC validation.
- Acknowledgement and error processing.
- Overlapping acknowledgement/error rejection.
- `maxEvents`, `moreAvailable` and short-polling behavior.
- JWS-signed poll response generation.
- Completion HMAC verification.
- Completion persistence.
- Endpoint-to-handler-to-service delegation.
- Subscription-scoped DAO updates and database-dialect behavior.


docs/event-notification-guide.md
```diff
@@ -34,39 +34,13 @@ administrative scopes.
 | Poll event deliveries | `notifications:events:poll` |
 | Submit delivery completion | `notifications:event-deliveries:complete` |
 
-### Configure publisher and receiver roles
+### Role assignment
 
-For application-to-application testing, create separate least-privilege roles
```
hasithakn, line 39: we need this section too right?

docs/event-notification-guide.md
```diff
@@ -102,7 +76,114 @@ RECEIVER_ACCESS_TOKEN="<event-receiver-access-token>"
...
+curl -k -X POST "${API_BASE}/events/poll" \
+  -H "Authorization: Bearer ${RECEIVER_ACCESS_TOKEN}" \
+  -H "Content-Type: application/json" \
+  -H "group-id: processor-1" \
+  -H "subscription-id: ${POLL_SUBSCRIPTION_ID}" \
+  -H "event-signature: ${POLL_SIGNATURE}" \
+  -d "${POLL_BODY}"
+```
+
+The response contains `sets`, keyed by `deliveryId`, and `moreAvailable`. Each
```
hasithakn, line 140: lets add sample response payload for both APIs

docs/event-notification-guide.md
```diff
@@ -224,30 +305,52 @@ these headers:
...
   "payload": {
-    "consentId": "c7c6b814-ef76-4eb4-9494-8af1b98a9ed1",
-    "status": "REVOKED"
+    "deliveryId": "<delivery-id>",
+    "eventId": "<event-id>",
+    "subscriptionId": "<subscription-id>",
+    "orgId": "example.com",
+    "groupId": "example-group",
+    "topic": "consent-status-changed",
+    "payload": {
```
hasithakn: there are 2 nested `payload` sections right? that might be confusing, shall we rename those, ex `eventPayload` or `responsePayload` etc?

dpdp-accelerator/components/org.wso2.dpdp.accelerator.event.notifications.service/src/main/java/org/wso2/dpdp/accelerator/event/notifications/service/impl/EventPublishServiceImpl.java
```diff
@@ -91,11 +112,274 @@ public EventPublishServiceImpl(EventDAO eventDAO, TopicDAO topicDAO, EventFanOut
...
+            throw new EventNotificationException(EventNotificationServiceConstants.ERROR_CODE_DELIVERY_NOT_FOUND,
+                    EventNotificationServiceConstants.ERROR_TITLE_DELIVERY_NOT_FOUND,
+                    EventNotificationServiceConstants.DELIVERY_NOT_FOUND_ERROR_MSG, 404);
+        }
+        if (!DeliveryStatus.DELIVERED.getValue().equalsIgnoreCase(delivery.get().getStatus())) {
+            throw new EventNotificationException(EventNotificationServiceConstants.ERROR_CODE_INVALID_STATE,
+                    EventNotificationServiceConstants.ERROR_TITLE_INVALID_STATE,
+                    EventNotificationServiceConstants.DELIVERY_COMPLETION_INVALID_STATE_ERROR_MSG, 409);
+        }
+        if (!HmacSigner.verifyCompletion(subscription.get().getSharedSecret(), safeDeliveryId,
```
hasithakn: ▎ Security: signature verification runs too late in completeDelivery()
▎
▎ Right now the order is: look up the delivery (line 246) → check it's in this group (line 253) → check it's delivered (line 260) → then verify the HMAC (line 265). That means a caller with no valid signature at all can still distinguish "delivery doesn't exist" (404) from "delivery exists but isn't delivered yet" (409) from "signature is wrong" (401) — three different outcomes leak whether a given deliveryId is real and what state it's in, all without ever proving they hold the shared secret.
▎
▎ The subscription (and its sharedSecret) is already resolved by line 254, right after the delivery lookup. Suggest moving the HMAC check to run immediately after that — right after line 258 — so an unauthenticated/incorrectly-signed request always just gets a flat 401 regardless of whether the delivery exists, matches the group, or has been delivered. pollEvents() already follows this order correctly (subscription resolved → HMAC verified → everything else), so this would just bring completeDelivery() in line with the same pattern.

hasithakn (approved):

PR #53, opened 2026-08-30
Implement event publishing for defined five events

hasithakn:
## Summary

Wires the 5 DPDP-relevant lifecycle events — consent updated, consent revoked,
consent expired, user data changed, user account deleted — into the existing
Event Notification Framework. The topic catalog (`DefaultTopic`) already
existed and was auto-provisioned per tenant, but nothing published to it; this
closes that gap end-to-end.

- New `DPDPLifecycleEventListener` interface (`event.notifications.common`) —
  a typed, DPDP-internal capture point with one method per event.
- `DPDPLifecycleEventPublisher` (`event.notifications.service`) implements it
  and fans each call out to `EventPublishService.publishEvent`, one per
  `DefaultTopic`. Registered as an OSGi service, gated by a new
  `[dpdp_accelerator.event_notifications.lifecycle_events] publishing_enabled`
  flag (default `true`).
- `identity.extensions` calls the listener from the 3 existing consent hooks
  (`DPDPConsentManagementListener`, renamed from `DPDPConsentHistoryListener`
  since it's no longer just history capture; `DPDPConsentExpiryReconciler`)
  and a new `DPDPUserLifecycleEventHandler` (`AbstractEventHandler`) for the 2
  user-side events, subscribed via `POST_DELETE_USER` / `POST_SET_USER_CLAIMS`.
- The 3 consent events publish real `purposes` resolved from the consent's
  `Receipt`; the 2 user events publish none — a user isn't tied to a single
  processing purpose, so this is a deliberate asymmetry, not a gap (see
  `docs/event-notification-guide.md`'s new "Has a purpose?" column).
- `[[event_handler]]` in `deployment.toml` registers the new handler with IS's
  Identity Event Framework — a stock IS 7.3.0 mechanism (same shape as
  `[[resource.access_control]]`), so no custom `identity-event.properties`
  file or `configure.sh` changes were needed.
- All lifecycle-event notification failures are caught and logged at the call
  site — never allowed to block the underlying consent/user mutation.
- Reused the existing shared `LogSanitizer.sanitize()` (`common` module)
  instead of adding another private per-class copy.

## Why

Subscribers (webhooks) can already register against these 5 topics, but
without a publisher nothing ever fired — the feature was half-wired. This
completes it so consent status changes and user data/account lifecycle events
actually reach the notification pipeline.

## Test plan

- [x] `mvn clean install` — full reactor build passes.
- [x] New unit tests: `DPDPLifecycleEventPublisherTest`,
      `DPDPUserLifecycleEventHandlerTest`, `DPDPLifecycleEventUtilTest`,
      `DPDPConsentSnapshotBuilderTest#resolvePurposes`, plus updated
      `DPDPConsentManagementListenerTest`.
- [x] Verified `[[event_handler]]` renders correctly against a real IS 7.3.0
      `identity-event.properties.j2` (decompiled and checked module
      auto-numbering).
- [ ] E2E: confirm a subscribed webhook actually receives a payload for each
      of the 5 topics on a live IS.


dpdp-accelerator/components/org.wso2.dpdp.accelerator.event.notifications.service/src/main/java/org/wso2/dpdp/accelerator/event/notifications/service/listener/DPDPLifecycleEventPublisher.java
```diff
@@ -0,0 +1,98 @@
...
+
+    @Override
+    public void onConsentUpdated(String orgId, String consentId, String previousStatus, String currentStatus,
+            String actionBy, List<String> purposes) {
+
+        Map<String, Object> payload = new HashMap<>();
+        payload.put("consentId", consentId);
+        payload.put("previousStatus", previousStatus);
+        payload.put("currentStatus", currentStatus);
+        payload.put("actionBy", actionBy);
```
ThiwankaChanditha: lets remove action by

ThiwankaChanditha (approved):

PR #54, opened 2026-08-30
Fix consent history gaps and improve the UI for latest IS consent API changes

hasithakn:
Description

- Implement latest IS consent API changes in the UI
- Fix postUpdateConsent silently dropping real state transitions from status-history (e.g. an EXPIRED consent revived to ACTIVE via expiry renewal).
- Show the resulting consent state in each expanded snapshot history entry (frontend).
- Consent list APIs (self + admin) now fetch purposes/authorizations inline instead of per-row lookups, and support relation/filter query params.
- Let a listed authorizer approve/reject a consent from the admin view, not just the subject from self-service.
- Generalize merge.sh's stale-webapp cleanup into one loop over every shipped webapp, instead of a hardcoded block per webapp.
- Fix CH-00002 (Forbidden - not the owner) incorrectly blocking a delegated consent's authorizer (e.g. a guardian) from viewing its status-history/full-history — the ownership check now also allows listed authorizers, not just the subject.


hasithakn:
image

Tested locally

PR #58, opened 2026-08-31
CI: supply update_tool_setup.sh instead of assuming the IS pack ships it

anjuchamantha:
## The problem

The E2E job's "Apply WSO2 U2 updates" step assumed `bin/update_tool_setup.sh` already exists in the WSO2 IS pack downloaded from GitHub releases. It doesn't — that script ships in WSO2's subscription-obtained product packs, not in the plain public release zip this workflow downloads (`https://github.com/wso2/product-is/releases/...`). This is why every E2E run that got past the label gate has been failing with:

```
bash: bin/update_tool_setup.sh: No such file or directory
Process completed with exit code 127
```

Confirmed by downloading and inspecting the actual release artifact: it contains `updates/product.txt` (the one file the script reads) but not the script itself.

## What this does

Adds an encrypted copy of `update_tool_setup.sh` under `.github/ci-tools/update_tool_setup.sh.enc` (AES-256-CBC, via `openssl`), and updates the "Apply WSO2 U2 updates" step to decrypt it into the downloaded pack's `bin/` before invoking it, using a new repository secret: `WSO2_UPDATE_TOOL_SETUP_DECRYPTION_PASSPHRASE`.

Verified locally end to end: the decrypted copy is byte-identical to the original script, and running it for real resolves and downloads a working update-tool binary for the target OS/architecture.

## What this doesn't fix

This only fixes the missing-script failure. It doesn't address the `SOLUTIONS_BOT_EMAIL` / `SOLUTIONS_BOT_U2_PASSWORD` secrets themselves — those still need to be configured as repo secrets by someone with admin access, same as before, for the release+U2 path to actually apply updates once this step gets past the part that was failing. `WSO2_UPDATE_TOOL_SETUP_DECRYPTION_PASSPHRASE` will need to be added as a repo secret as well before this step can succeed in CI.


anjuchamantha:
Closing this one.

PR #63, opened 2026-08-31
E2E tests for Complaints, Event Notification and Consent History

ParameswaranSajeenthiran:
---                                                                                                                                                                                
                                                                                                                                                                                     
  Purpose                                                                                                                                                                            
                                                                                                                                                                                     
  The Event Notification Framework, Complaint Management, and Consent History features had no automated end-to-end coverage — regressions in any of them could only be caught by     
  manual testing. This PR adds Playwright integration test suites for all three, driven against a real, already-running WSO2 IS + accelerator deployment.                            
                                                                                                                                                                                     
  Goals                                                                                                                                                                              
                                                                                                                                                                                     
  - Full E2E coverage for Complaint Management: data-principal complaint lifecycle, officer workflows, threaded replies, attachments, authorization/isolation, both through the UI   
    (tests/07-complaints) and the REST API directly (tests/06-complaints-api).                                                                                                       
  - Full E2E coverage for Consent History: the lifecycle timeline and full-history dialog on both the self (/consents/:id) and admin (/administration/consents/:id) detail pages.    
  - Full E2E coverage for the Event Notification Framework: Topics, Subscriptions, Events, real signed webhook delivery (HMAC-SHA256, retries, backoff), authorization, and tenant   
    isolation (tests/08-event-notifications).                                                                                                                                        
  - Wire the Event Notification webhook-delivery tests into CI with no secrets required — the e2e workflow now patches the freshly-deployed IS instance to allow the runner's own    
    private IP as a webhook callback target before the server starts, so real signed webhook delivery is exercised on every CI run. An optional NGROK_AUTH_TOKEN secret is also      
    supported as a higher-fidelity alternative (public HTTPS tunnel).
    
    Approach                                                                                                                                                                           
                                                                                                                                                                                     
  - New API clients (ComplaintApiClient.ts, EventNotificationApiClient.ts) and ~20 new page objects, following this suite's existing pattern (ConsentApiClient.ts, locators live only
    in pages/, personas resolved through fixtures/auth.fixtures.ts).                                                                                                                 
  - utils/webhookReceiver.ts: a real local HTTP receiver used by the webhook-delivery tests, supporting two ways to become network-reachable by the IS host (loopback callback URLs  
    are rejected unconditionally by EventNotificationUrlValidator, so this can't just bind to localhost):                                                                            
    - An ngrok tunnel (NGROK_AUTH_TOKEN) — genuine public HTTPS, no deployment config change needed.                                                                                 
    - A LAN address (WEBHOOK_RECEIVER_HOST) for local dev, requiring allow_private_network_callback_targets = true on the target deployment.                                         
    - Binds within 8443–8455 rather than an OS-assigned ephemeral port, since the accelerator's own allowed_callback_ports allow-list would otherwise reject it.                     
  - .github/workflows/pr-checks.yml: the e2e job patches the freshly-generated deployment.toml (private-network flag + widened port range) and computes the runner's own IP, before  
    the server starts — no restart problem, since nothing has started yet on a fresh CI runner.                                                                                      
  - A handful of real product bugs were found while writing these tests and documented (not fixed, per this repo's separation of test/product concerns) in                           
    tests/08-event-notifications/README.md's "Bugs found" section — notably that SubscriptionHandler.createSubscription silently ignores a caller-supplied groupId and always forces 
    it to the org id, and that EventEndpoint.listEvents hardcodes the org id as the GROUP_ID filter on every call regardless of what's requested.                                    
  - Tests that need infrastructure this suite genuinely can't provide (DB fault injection for transaction-rollback testing, forcing a stuck in-flight delivery) are test.skip() with 
    a documented reason rather than omitted silently.   
    
    
    User stories                                                                                                                                                                       
                                                                                                                                                                                     
  - As a maintainer, PR changes to complaint management, consent history, or event notifications are now caught by CI before merge.                                                  
  - As a reviewer, I can see webhook delivery — including HMAC signature verification and retry/backoff behavior — actually exercised in CI, not just unit-tested.                   
                                                                                                                                                                                     
  Release note                                                                                                                                                                       
                                                                                                                                                                                     
  Adds automated end-to-end test coverage (Playwright) for Complaint Management, Consent History, and the Event Notification Framework, including real signed webhook delivery in CI.
                                                                                                                                                                                     
  Documentation                                                                                                                                                                      
                                                                                                                                                                                     
  N/A — test-suite-only change, no product documentation impact.                                                                                                                     
                                                                                                                                                                                     
  ---


.github/workflows/pr-checks.yml
```diff
@@ -268,6 +257,39 @@ jobs:
           bash bin/merge.sh     "$IS_HOME"
           bash bin/configure.sh "$IS_HOME"
 
+      # CI-only, no-secret path for tests/08-event-notifications' webhook-delivery tests: unlike a
```
anjuchamantha: This is used for a test right? Can't we handle it inside the test suite itself?

.github/workflows/pr-checks.yml
```diff
@@ -341,6 +369,9 @@ jobs:
           TEST_CONSENT_ADMIN_PASSWORD=${TEST_PASSWORD}
           TEST_USER_2_USERNAME=${{ steps.users.outputs.user2 }}
           TEST_USER_2_PASSWORD=${TEST_PASSWORD}
+          NGROK_AUTH_TOKEN=${{ secrets.NGROK_AUTH_TOKEN }}
```
anjuchamantha: Why do we need ngrok?

.github/workflows/pr-checks.yml
ParameswaranSajeenthiran: yes, do you mean to move this script to the tests folder or handle it by a node library?

.github/workflows/pr-checks.yml
ParameswaranSajeenthiran: For event notifications, we need to configure a subscriber with a webhook URI. Initially, I planned to use a hosted webhook endpoint secured with an AUTH_TOKEN. However, later I decided to go with a local deployment approach instead.

As a result, the hosted webhook configuration is no longer required and has become an orphaned configuration. I removed this unused configuration in the following commits https://github.com/wso2/dpdp-accelerator/pull/63/commits/6ab8d36a541a3309171fef12814329cfbc35898b and https://github.com/wso2/dpdp-accelerator/pull/63/commits/14f2e0df6937e42f8e0cf4002a80d7d3b715eae6

Issue #81, opened 2026-09-01
Frontend: fix the out-of-sync package-lock.json and the 23 ESLint violations, then enable npm ci and lint in CI


anjuchamantha:
The CI restructure adds a `Frontend quality` job that currently enforces only Prettier formatting. Two stricter checks were deliberately left out because the tree cannot pass them today, and a permanently-red check is worse than no check. Both should be enabled in the change that fixes them.

## 1. `package-lock.json` is out of sync, so `npm ci` cannot run

`dpdp-accelerator/react-apps/consent-portal/frontend/package-lock.json` contains **no `esbuild` entries at all**, even though `vite@7.3.5` depends on `esbuild@^0.27.0` and `package.json` carries a security override pinning it:

```json
"esbuild@>=0.27.3 <0.28.1": "0.28.1"
```

`npm ci` therefore fails outright:

```
npm error `npm ci` can only install packages when your package.json and package-lock.json are in sync.
npm error Missing: esbuild@0.28.1 from lock file
npm error Missing: @esbuild/aix-ppc64@0.28.1 from lock file
   ... (all 26 platform variants)
```

The Maven build masks this because it runs `npm install`, which silently repairs the tree in place (and rewrites the lockfile as a side effect on every build).

Notes for whoever picks this up:
- `dpdp-integration-test-suite/package-lock.json` is healthy — `npm ci` passes there. Only the frontend lockfile is affected.
- Regenerating it needs Node **22.12+** per the repo's own prerequisite. An attempt on Node 22.8.0 produced a ~5000-line re-serialisation and still could not satisfy `npm ci` (the override's upper bound `<0.28.1` mapping to `0.28.1` appears to resolve differently between `npm install` and `npm ci`), so it needs doing on a supported Node with the result actually verified by running `npm ci`.
- Once `npm ci` passes, change the `Install` step in `.github/workflows/pr-build.yml` from `npm install --no-audit --no-fund` to `npm ci`. That also makes CI fail on future `package.json`/lockfile drift, which nothing catches today.

## 2. `npm run lint` has 23 pre-existing violations

`npm run lint` currently reports 74 errors. 51 were Prettier issues, already fixed by the formatting commit in the CI PR. The remaining 23 need real code changes:

| Count | Rule |
| --- | --- |
| 8 | `react/require-default-props` |
| 5 | `class-methods-use-this` (all in `vitest.setup.ts`'s Worker stub) |
| 3 | `no-void` |
[...]

Issue #80, opened 2026-09-01
Repo-admin tasks to enforce the restructured CI (branch ruleset, code-scanning pre-check)


anjuchamantha:
The upcoming CI restructure (splitting `pr-checks.yml` into `pr-build.yml` / `pr-e2e-gate.yml` / `pr-e2e.yml`, reporting the E2E result as a single commit status, adding frontend quality checks, dependency review and CodeQL) is designed to need **no repository settings** — but its enforcement stays *advisory* until a repo admin completes the items below. Today a red or pending check only tells a reviewer not to merge; the Merge button still works.

## 1. Required: create a branch ruleset (turns CI advisory → enforced)

**Settings → Rules → Rulesets → New branch ruleset**, targeting `main` (and `dev` if desired), with **Require status checks to pass** listing:

| Required check | Kind |
| --- | --- |
| `Build` | check run (from `pr-build.yml`) |
| `Frontend quality` | check run (from `pr-build.yml`) |
| `Dependency review` | check run (from `pr-build.yml`) |
| `E2E (label-gated)` | **commit status context** (posted by `pr-e2e-gate.yml` / `pr-e2e.yml`) — not a check run |

Why `E2E (label-gated)` works as a required check: it is set to *pending* on every push and only flips green after a maintainer applies the `Action/trigger-e2e` label and the E2E suite passes. A pending status blocks merge, so no PR can merge without E2E having run on its final commit. (This replaces the old `E2E integration tests (required)` workaround job — GitHub treats a *skipped* required job as passing, which is why that job existed; commit statuses have no such loophole.)

## 2. Pre-check before the CodeQL workflow lands

**Settings → Code security → Code scanning**: confirm **Default setup is OFF**. If default setup is enabled, the advanced `codeql.yml` workflow's SARIF uploads are rejected with a "default setup enabled" error. If it is already off (no scanning is configured today), nothing to do.

## 3. Optional, recommended

- Enable **Dependabot alerts** and **Dependabot security updates** (Settings → Code security). The `dependabot.yml` *version updates* need no admin action — committing the file is enough.
- Once CodeQL runs are stable, add its `Analyze (java-kotlin)` / `Analyze (javascript-typescript)` checks to the ruleset as well, if scanning should block merges rather than just report.

---

The check/status names above land with the CI restructure PR; the ruleset can be created as soon as that PR is merged. Nothing in the workflows needs to change when the ruleset is added.

anjuchamantha:
[...]

PR #82, opened 2026-09-02
Restructure the CI workflows: remove the skipped-check noise, test the merge result, gate on a single status

### LINKED ISSUE #78: [Epic] Revise & Cleanup Git Action workflows

anjuchamantha:
## Why

Every PR to this repository accumulates a wall of check runs, most of them meaningless, and one of them permanently red. On PR #69's head commit there were **12 check runs: 6 skipped, 1 failed**. The failure was `E2E integration tests (required)` reporting on the run from *before* the E2E label was applied — nothing ever cleared it, so a green PR looked red.

The cause is that `pr-checks.yml` declared both `pull_request` and `pull_request_target`. Every push fired two workflow runs (three once the label was applied), and each run reported *all four* of its jobs — the ones its own event had disabled reporting as `skipped`.

Fixing that turned up several other things worth correcting, listed below.

## What changed

**The trigger split.** `pr-checks.yml` is replaced by three single-trigger workflows, so no job ever reports on an event it does not belong to:

| Workflow | Trigger | Jobs |
| --- | --- | --- |
| `pr-build.yml` | `pull_request` | `Build`, `Frontend quality` |
| `pr-e2e-gate.yml` | `pull_request_target` (push/open/reopen) | `Reset E2E label on push` |
| `pr-e2e.yml` | `pull_request_target` (labeled), `workflow_dispatch` | `e2e`, `Report E2E result` |

**The E2E result is now a commit status, not a check run.** A status is keyed by its context string, so writing `E2E (label-gated)` again *replaces* what the PR shows: pending on every push, success or failure once a labelled run finishes. Exactly one always-current entry per commit.

This also removes the reason the old `e2e-required` job existed. Rulesets treat a *skipped* required check as satisfied, so the `e2e` job itself could never have been the gate. A pending status has no such loophole, which makes `E2E (label-gated)` directly usable as a required status check (see #80).

**E2E tests the merge result.** It checked out the PR's head SHA — the branch in isolation, which can be green while still breaking `main` on merge. It now checks out `refs/pull/N/merge`. Because GitHub recomputes that ref whenever the PR moves, `e2e.yml` verifies its second parent is still the head commit the label authorised and fails closed otherwise. That needs `fetch-depth: 2`; verified that `git rev-parse HEAD^2` fails outright on a depth-1 clone of a merge commit.

**Two concurrency bugs.** The group was `${{ github.workflow }}-${{ github.ref }}`, and under `pull_request_target` `github.ref` is the *base* branch — so every PR targeting `main` shared one `cancel-in-progress` group and labelling one PR cancelled another's in-flight E2E. Both PR workflows now key on the PR number. Separately, because `labeled` cannot be filtered by label name in `on:`, a run created by *any other* label would join that group and cancel a legitimate E2E before skipping its own jobs; those runs now get a throwaway group.

**PR runs no longer receive subscription secrets.** `secrets: inherit` handed every PR run `SOLUTIONS_BOT_EMAIL` / `SOLUTIONS_BOT_U2_PASSWORD` / `WSO2_UPDATE_TOOL_SETUP_DECRYPTION_PASSPHRASE`. A labelled PR run always builds product-is from master, and every consumer of those secrets is gated on `is_source == 'release'` — so the one path that executes possibly-fork-authored code was the one carrying credentials it could not use. They are now passed only on a manual dispatch. The label gate is still needed: `pull_request_target` runs with the base repository's write-capable token and cache scope.

**The frontend Vitest suite now runs in CI.** The consent-portal module ran `npm install` and `npm run build`, and that build script chains `tsc`, `security:verify` and `i18n:verify` — but never `npm test`. All 297 frontend unit tests could go red without a single check noticing, and `release-builder.yml`'s claim that `mvn clean install` is the full gate was untrue for the frontend. It is bound to the Maven `test` phase, so the PR build, the E2E job's build and the release build all gate on it, and `-DskipTests` skips it alongside surefire.

**Prettier is enforced.** It had never run in CI, so the tree had drifted from its own `.prettierrc` — 17 files, fixed here in a separate mechanical commit.

**Hardening.** Every action is pinned to a commit SHA with its version in a trailing comment (a mutable tag is an arbitrary-code-execution vector, worst under `pull_request_target`); Dependabot keeps the pins fresh across four ecosystems; CodeQL scans Java and TypeScript on PRs, pushes and weekly.

Also corrected: comments claiming PR runs used the `release`+U2 path, which they never did — the `is_source` fallback has always resolved to `master`.

## Verified

Behaviour was exercised on a real PR in a fork, not just reasoned about:

- **The check set is clean** — 6 checks, **zero skipped**, one pending `E2E (label-gated)`.
- **Full E2E passes** — merge-ref checkout, guard log `Merge ref second parent matches the labelled head 5b2108ca…`, accelerator build (Vitest 42 files), IS deployed from product-is master, **Playwright 149 passed in 11.9m**, and the gate flipping to a single green context.
- **An unrelated label no longer kills a run** — applying `documentation` mid-E2E produced a skipped no-op run while the real E2E kept going.
- **Pushing strips the label and re-pends the gate**, and cancels an in-flight E2E; the superseded commit keeps its pending status rather than receiving a stale failure.
- `mvn -B -ntp clean install` green on the full reactor; `actionlint` clean on all six workflows; all nine pinned SHAs verified to exist and to match their version comments.

## Deliberately left out

Three checks are *not* enabled here, because each would have been red on arrival and a permanently-red check is exactly the noise this PR removes:

- **`npm run lint`** — 23 pre-existing ESLint violations needing real code changes (#81).
- **`npm ci`** — `package-lock.json` is out of sync: it contains no `esbuild` entries at all, though `vite@7.3.5` requires it and `package.json` overrides it to 0.28.1 (#81). The Maven build masks this by using `npm install`.
- **`Dependency review`** — the action fails with *"Dependency review is not supported on this repository"*; this repository's dependency graph is off (`/dependency-graph/sbom` returns 404 here, while it returns an SBOM for `actions/checkout`, `thunder-id/thunderid` and `wso2/product-is`). Re-adding it is a copy-paste once an admin enables the toggle — YAML is in #80.

Closes #78.

## Follow-ups

- **#80** — repo-admin steps: create the ruleset requiring `Build`, `Frontend quality` and the `E2E (label-gated)` status; enable the dependency graph; keep CodeQL *default setup* disabled, since this is advanced setup and the two are mutually exclusive.
- **#81** — fix the lockfile and the ESLint violations, then enable `npm ci` and `npm run lint`.

Both stay open after this merges: neither is done by it. #80 needs repository settings nobody on this PR can change, and until those are applied everything here is advisory rather than enforced; #81 tracks the three checks deliberately left out.

Until the ruleset exists, all of this is advisory: a red or pending `E2E (label-gated)` status tells a reviewer not to merge, but does not block the button.

> One sequencing note for #80: `pull_request_target` workflows load from the base branch, so no `opened` event replays for PRs that are already open when this merges. Those PRs will carry no `E2E (label-gated)` status until their next push — worth letting them cycle before turning the ruleset on.

---

**Update since opening:** the commit status was renamed from `E2E gate` to **`E2E (label-gated)`**, and its descriptions now name the action to take in every state a reader has to act on - `E2E has not run on this commit - a maintainer applies Action/trigger-e2e to run it` when pending, and `E2E failure - see Details, then re-apply Action/trigger-e2e to run it again` on failure. Clicking **Details** on the pending status now lands on a rendered explanation of the label process rather than a bare API call, so a first-time contributor is not left guessing. The two PR-visible jobs were renamed to say what they do (`Reset E2E label on push`, `Report E2E result`), following FS Accelerator's own `Reset-integration-label-on-push`.

A revert commit also removes an automatic `1.0.1-SNAPSHOT` version bump that reached this branch by accident: a manual Release builder dispatch on the fork published a test release, and its post-release job pushed the next-development-version bump to the fork's `main`, which this PR is raised from. The reactor version is unchanged from `main`.

Fixes https://github.com/wso2/dpdp-accelerator/issues/78


anjuchamantha:
### Reading the checks on this PR

This PR is judged partly by the workflows it replaces, which makes its own check list misleading. `pull_request_target` workflows always load from the **base branch**, so `pr-checks.yml` on `main` is still the one running for that trigger here — the new `pr-e2e-gate.yml` / `pr-e2e.yml` do not exist on `main` yet and cannot run on this PR.

What each entry actually comes from:

| Check | Source | Note |
| --- | --- | --- |
| `Build`, `Frontend quality` | **new** `pr-build.yml` | `pull_request`, so it loads from this PR — this is the new behaviour, and it is green |
| `Analyze (java-kotlin)`, `Analyze (javascript-typescript)` | **new** `codeql.yml` | same, green |
| `E2E integration tests (required)` — ❌ | **old** `pr-checks.yml` on `main` | the very check this PR removes |
| `e2e`, `Reset E2E label on push`, second `Build` — skipped | **old** `pr-checks.yml` on `main` | the skipped-check noise this PR removes |
| no `E2E (label-gated)` status | — | `pr-e2e-gate.yml` is not on `main` yet, so nothing publishes it here |

So the red ❌ and the three skipped entries are produced by the code being deleted, not by the code being added. They are the last time they will appear: the fork verification below shows what a PR looks like once this is on `main`.

To turn the ❌ green, apply `Action/trigger-e2e` as usual — that runs the **old** E2E path from `main`, which is the correct gate for this PR.

### What the new behaviour looks like

Verified on a real PR in a fork whose `main` carried these workflows:

```
Build                             success
Frontend quality                  success
Analyze (java-kotlin)             success
Analyze (javascript-typescript)   success
Reset E2E label on push           success
E2E (label-gated)                 pending → success   ← one entry, overwritten in place
```

Zero skipped, no stale red. The full E2E passed there end to end: merge-ref checkout, the guard logging `Merge ref second parent matches the labelled head 5b2108ca…`, the accelerator build running Vitest (42 files), IS deployed from product-is master, and **Playwright 149 passed in 11.9m**.

hasithakn (approved):

PR #142, opened 2026-09-04
Fix Stale Reads on MySQL in Event Notifications DAO

### LINKED ISSUE #107: # Event-notification subscription list intermittently returns empty after creating subscriptions
## Summary

Right after creating event-notification webhook subscriptions, `GET /api/dpdp/event-notifications/v1/subscriptions` correctly returns them. Reloading and re-issuing the exact same request (same tenant, same query params) then returns `{"items":[],"total":0}` — an empty result from the server itself, confirmed via the browser network tab, not a frontend display issue.

## Example

First request:
```json
{"items":[{"subscriptionId":"b1138476-...","orgId":"t-a.com","topic":"consent.revoke","status":"active",...},{"subscriptionId":"febac1d2-...","orgId":"t-a.com","topic":"consent.expire","status":"active",...}],"total":2}
```
Same request again shortly after:
{"items":[],"total":0}

Root cause

JDBCPersistenceManager.getDBConnection() always sets connection.setAutoCommit(false). Every read method in SubscriptionDAOImpl — including listSubscriptions (SubscriptionDAOImpl.java:389-443) and getSubscriptionById (:211-219) — only calls DatabaseUtils.closeConnection(conn) in its finally block. It never calls commitTransaction or rollbackTransaction after the SELECT.

That leaves each read connection's transaction open when it goes back to the pool. Under MySQL InnoDB's default REPEATABLE READ isolation, a connection whose previous transaction was never closed out can be handed back out on a later request still holding its original snapshot — so a subsequent list request that happens to reuse that same pooled connection can see the database as it looked before the new subscriptions were inserted, even though they were already committed by the write path (SubscriptionDAOImpl.addSubscription does correctly call commitTransaction, confirmed at line 75).

This is a systemic gap, not specific to subscriptions — the same missing-commit-on-read pattern shows up elsewhere in the event-notifications DAO layer (e.g. topic/delivery reads), so other list/read endpoints in this module are likely exposed to the same intermittent staleness.

Ruled out

- No in-memory cache anywhere in event.notifications.*.
- No cleanup/expiry job touches SUBSCRIPTION rows (the only background job is webhook-verification, which flips PENDING→ACTIVE/STALE and never deletes/hides rows, and only applies to webhook-mode subscriptions).
- No group_id/actor/"created by" filter exists in the list query's WHERE clause, so this isn't a scoping/identity mismatch.
- Not a pagination bug — the count query also returned 0, and the offset/limit math is identical between the two calls.
- Confirmed unrelated to the separately-diagnosed complaint-reply frontend bug — different layers (backend SQL/transaction state vs. frontend query cache), no shared code.

Suggested fix

Every read-only DAO method that opens a connection via getDBConnection() should call DatabaseUtils.commitTransaction(conn) (or explicitly use a read-only/auto-commit connection) before closeConnection, the same way the write paths already do — so a connection is never returned to the pool mid-transaction. Worth auditing the rest of the event-notifications DAO layer (topics, deliveries) for the same gap while fixing this.

Where

- SubscriptionDAOImpl.java:389-443 (listSubscriptions)
- SubscriptionDAOImpl.java:211-219 (getSubscriptionById)
- JDBCPersistenceManager.java:90-99 (connection setup, setAutoCommit(false))

ThiwankaChanditha:
## Summary

This PR fixes stale reads in the Event Notification framework when using MySQL with InnoDB and the default `REPEATABLE READ` isolation level. This Fixes #107.

The issue occurred because Event Notification DAO methods acquired database connections with `autoCommit=false` but did not consistently complete the transaction before returning the connections to the Tomcat JDBC pool. When a pooled connection was reused, MySQL could continue using the old MVCC snapshot. As a result, data committed by a previous request was not visible to subsequent reads.

This caused symptoms such as:

- Newly created topics or subscriptions not appearing in list responses.
- Incorrect pagination totals, for example `total: 10` instead of `total: 11`.
- `GET` requests returning `EN-4040` for resources that had just been created.
- Different concurrent requests returning different database results depending on which pooled connection they received.

### Centralized transaction management

A new transaction wrapper was added to `DatabaseUtils`:

```java
DatabaseUtils.executeInTransaction(connection -> {
    // transactional operation
});
```

The wrapper is responsible for:

1. Acquiring a connection with auto-commit disabled.
2. Passing the connection to the transactional operation.
3. Committing when the operation completes successfully.
4. Rolling back when the operation or commit fails.
5. Closing the connection in all cases.
6. Propagating commit failures so callers do not receive a successful response when the transaction was not persisted.

The implementation also preserves the original failure when rollback or connection cleanup fails.

### Connection lifecycle hardening

`JDBCPersistenceManager.getDBConnection()` was updated to close the connection if configuring `autoCommit=false` fails. Any cleanup failure is attached to the original exception as a suppressed exception.

This prevents partially initialized connections from being returned to the pool.

### DAO changes

The Event Notification DAOs were changed to use connections supplied by their callers. Transaction lifecycle management is no longer hidden inside individual DAO operations.

The following DAO areas were migrated:

- `TopicDAO`
- `SubscriptionDAO`
- `EventDAO`
- `DeliveryDAO`
- `DeliveryAckDAO`

The affected operations include:

- Topic lookup, listing, creation, and deregistration
- Subscription lookup and listing
- Subscription purpose lookup
- Active-subscription checks
- Subscription recovery queries
- Event lookup and search
- Event-purpose lookup
- Delivery lookup
- Poll-delivery retrieval
- Webhook dispatch-context loading
- Delivery history queries
- Subscription delivery queries
- Organization-level delivery queries
- Delivery acknowledgement lookup

This ensures that related database operations can execute within one service-owned transaction and that every connection is completed before it is returned to the pool.

### Service-layer migration

The service implementations were updated to establish the transaction boundary around the complete business operation rather than around isolated DAO calls.

The migrated flows include:

- Topic creation and deletion
- Topic listing and lookup
- System-topic reconciliation
- Subscription creation
- Subscription verification and retry
- Subscription deletion
- Subscription listing and lookup
- Event publication
- Event and delivery queries
- Poll delivery retrieval
- Poll delivery acknowledgement
- Delivery history retrieval
- Webhook completion processing
- Delivery recovery
- Stuck-delivery recovery
- Manual event retrigger processing

This also preserves consistency for operations that perform multiple reads and writes, such as event publication and delivery creation.

### Webhook and recovery processing

The webhook worker and delivery task were updated to use explicit transaction boundaries for:

- Claiming pending deliveries
- Claiming stuck `in_flight` deliveries
- Recording successful delivery
- Recording retryable failures
- Recording permanent failures
- Releasing deliveries
- Updating retry timestamps
- Recording completion information
- Performing manual delivery retries

The recovery service was also migrated so its delivery and subscription recovery queries use properly scoped transactions.

## Test updates

Tests were updated to reflect the new connection-aware DAO APIs and service-owned transaction boundaries.

Coverage was added or updated for:

- Successful transaction commit
- Rollback after runtime failures
- Rollback after assertion errors
- Commit failure propagation
- Connection closure
- Connection cleanup when auto-commit configuration fails
- DAO read paths
- DAO write paths
- Event and acknowledgement operations
- Topic concurrency behavior
- Subscription concurrency behavior
- Delivery history consistency
- Poll delivery processing
- Webhook delivery state transitions
- Delivery recovery
- Manual event retrigger behavior
- Database dialect behavior and schema parity

## Validation

The DAO test suite completed with:

- 63 tests executed
- 0 failures
- 0 errors
- 2 skipped

A live MySQL verification was also performed.

Before the fix:

- Multiple connections were pre-warmed in the connection pool.
- A new subscription was created.
- Concurrent subscription-list requests returned inconsistent totals.
- Some requests continued returning the old count.
- Direct subscription lookup could return `EN-4040` for the newly created subscription.

After the fix:

- The same concurrent workflow consistently returned the updated subscription count.
- Individual subscription lookup requests returned `200 OK`.
- Newly committed topic, subscription, event, and delivery records became visible across subsequent pooled-connection requests.
- No stale connection snapshots were observed after read operations.

## Impact

The change applies to the Event Notification persistence and service flows. It does not change the public event-notification API contract, but it changes the internal transaction ownership model to ensure reliable visibility of committed data and consistent behavior with pooled MySQL connections.


dpdp-accelerator/components/org.wso2.dpdp.accelerator.event.notifications.service/src/main/java/org/wso2/dpdp/accelerator/event/notifications/service/impl/EventPublishServiceImpl.java
```diff
@@ -143,46 +143,22 @@ public EventPollingResponseDTO pollEvents(String orgId, String groupId, String s
         if (configurationService == null || signedEventPayloadFactory == null || subscriptionDAO == null) {
```
hasithakn, line 137: Shall we remove eventFanOutService and move the fanOutEvent method to [EventPublishServiceImpl.java‎]
 if possible we can make that a private method right?

hasithakn (approved):

PR #163, opened 2026-09-07
Update Admin "All Complaints" Queue View

### LINKED ISSUE #161: Admin View - Resolved complaints are not visible in the all status view.
When an admin logs into the portal, the complaint view does not show the resolved complaints, but the Status filter includes all statuses. Ideally, it should display all statuses.

Image

Tested with MySQL DB.

### LINKED ISSUE #201: DPO's Complaint Search Bar doesn't fetch from Backend
## Summary

The search box in the officer/admin complaint queue only filters complaints that are already loaded in the browser for the current page.

It does not query the backend, so searching for a complaint that is not on the currently displayed page returns no results, even though the complaint exists and would be found on another page.

## Steps to Reproduce

1. Log in as a complaint officer/admin and open the **Complaint Management** queue.
2. Ensure there are more complaints than can fit on a single page (`rowsPerPage`).
3. Enter the **reference ID** or **Data Principal name** of a complaint that exists on a different page.
4. Observe that the table shows no matching rows, even though the complaint exists.

## Expected Behavior

The search should query the full complaint set on the server, matching by **reference ID** or **Data Principal name**.

The API should return the correct paginated results regardless of which page the matching complaint would otherwise appear on.

## Actual Behavior

`ComplaintQueuePage.tsx` calls `useManagedComplaintListQuery` with only:

* `status`
* `priority`
* `limit`
* `offset`

The search term is never passed to the API.

Instead, the search is applied using a local `.filter()` on `pageComplaints`, which contains only the complaints already fetched for the current page.

As a result, the search can only narrow down results **within the currently loaded page** and cannot find complaints located on other pages.

ParameswaranSajeenthiran:
# Purpose

Two related defects in the officer/admin complaint queue (`ComplaintQueuePage.tsx`):

1. The "All" status filter silently excluded resolved complaints. `ComplaintQueuePage.tsx` removed them client-side, in the `rows` memo, after the server had already paginated the result set. Beyond the label being wrong, this broke the pager in ways an officer sees directly:
   - `total` is returned by the server unfiltered, so the pager read "1–10 of 80" while the table rendered six rows.
   - A page whose rows all happened to be resolved rendered empty, mid-list, with **Next** still enabled.
   - The "Resolved" stat tile immediately above the table counts resolved complaints server-side, so the tile and the list below it contradicted each other.
2. The search box never queried the backend at all. It filtered client-side over whatever page `useManagedComplaintListQuery` had already fetched (`status`/`priority`/`limit`/`offset` only — no search term was ever sent). Searching for a complaint's reference ID or Data Principal name found it only if it happened to already be on the currently loaded page; otherwise the table showed nothing, even though the complaint existed.

# Related Issues

Fixes [wso2/dpdp-accelerator#161](https://github.com/wso2/dpdp-accelerator/issues/161) — *Admin View - Resolved complaints are not visible in the all status view.* The admin's complaint view omitted resolved complaints even though the Status filter offers every status; reported against MySQL. This is the client-side `rows` memo described above, removed by this change.

Fixes [wso2/dpdp-accelerator#201](https://github.com/wso2/dpdp-accelerator/issues/201) — *DPO's Complaint Search Bar doesn't fetch from Backend.* The search box only narrowed the already-loaded page instead of querying the full complaint set server-side.

# Goals

Make "All" mean all, so the queue's pagination totals and row counts agree, and the stat tiles stop contradicting the table beneath them. Resolved complaints become visible in the default view without any new filter or control.

Make the search box search the entire complaint set on the server — by reference ID, Data Principal display name, or raw userId — regardless of which page a match would otherwise land on, matching what #201 asked for and going one field further (userId) for cases where no display name was ever resolved.

# Approach

## Resolved complaints hidden by "All"

Removed the `CLOSED_OUT_STATUSES` constant and the status branch inside the `rows` memo in `ComplaintQueuePage.tsx`; the memo now narrows only by the reference-id/name search (see below - now also server-side). Status and priority are already applied server-side via `useManagedComplaintListQuery`, so "All" sends no status parameter and the server's `total` matches what renders. The now-unused `ComplaintStatus` import was dropped.

Resolved complaints were never unreachable: "Resolved" is an existing option in the status dropdown and still works. What changes is the default view.

## Search bar now queries the backend

- **DAO** (`org.wso2.dpdp.accelerator.complaint.mgt.dao`): `ComplaintQueryBuilder` gained a `search` filter - a case-insensitive substring match (`LOWER(...) LIKE ? ESCAPE '!'`) against `REFERENCE_ID`, `USER_NAME`, or `USER_ID`, OR'd together. This is deliberately separate from the existing `userId` filter, which stays an **exact** match (see the OpenAPI description) for callers that want to scope to one specific Data Principal.
- **Shared utility**: the LIKE-escaping helpers (`escapeLikePattern`, `buildCaseInsensitiveContainsPattern`, `buildEscapedLikePredicate`) moved from being duplicated per-feature into `org.wso2.dpdp.accelerator.common.util.QueryBuilderUtils`, since `complaint.mgt.dao` is now a second real consumer alongside `event.notifications.dao`. The `event-notifications` module itself is untouched - it still has its own copy, unchanged, so this PR carries zero behavioral risk to that feature.
- **Service/endpoint**: `search` threaded through `ComplaintService`/`ComplaintServiceImpl`, `ComplaintHandler`, and `ComplaintEndpoint` (`GET /complaints?search=`) as a plain pass-through (no validation needed - it's free text). The `/me/complaints` surface intentionally does **not** get a `search` param - a Data Principal's own list is already scoped to their own `userId`, so there's nothing for a cross-user text search to narrow down.
- **OpenAPI**: added the `search` query parameter to `GET /complaints` in `complaint-server-API.yaml`, documented as a case-insensitive substring match, distinct from the existing exact-match `userId` parameter.
- **Frontend**: `ComplaintQueuePage.tsx` now passes the search box's value to `useManagedComplaintListQuery` (debounced 300ms, so typing doesn't fire a request - and a `LIKE` scan - on every keystroke) instead of filtering the already-fetched page. The old client-only substring filter is gone.

## UI

The queue table under the default filter now includes rows with the "Resolved" status chip. Typing in the search box now returns matches from the whole complaint set, not just the currently visible page. No new or changed user-facing text (the existing placeholder, "Search by reference ID or user", already described the intended behavior), so no documentation@wso2.com review is needed.

No screenshot attached - the deployment's database was cleared earlier in this session, so there is no complaint data to capture a meaningful before/after against. The screenshot on #161 shows the reported behaviour, and #201's repro steps describe the search defect.

# User Stories

- **As a DPO or consent admin**, when I open the complaint queue with no filter applied, I see every complaint including the ones already resolved, and the pagination count matches the rows in front of me.
- **As a DPO or consent admin**, when I search the queue by a complaint's reference ID or the Data Principal's name, I find it regardless of which page it would otherwise be on.

# Release Note

> Fixed the complaint queue's "All" status filter excluding resolved complaints, which caused the result count and pagination to disagree with the rows displayed.
>
> Fixed the complaint queue's search box only searching the currently loaded page instead of the full complaint set.

# Documentation

**N/A** - no configuration, API-contract, or documented behaviour changes beyond the OpenAPI spec's own description of the new `search` query parameter, which is not otherwise surfaced in `docs/`.

# Training

**N/A** - no training content covers the officer queue filter/search behaviour.

# Certification

**N/A** - this is a defect fix in a filter and a search box's default scope; it does not introduce a concept, configuration, or API that a certification question could test.

# Marketing

**N/A** - internal defect fix, no promotable feature.

# Automation Tests

## Unit Tests

**Resolved-complaints fix**: no unit-test change. `ComplaintQueuePage.tsx` has no Vitest coverage today (no test file references it), so there is no coverage delta to report.

**Search fix**: real Java-side coverage added/updated across four modules -
`org.wso2.dpdp.accelerator.common` (new `QueryBuilderUtilsTest` - the moved utility), `complaint.mgt.dao` (new `ComplaintQueryBuilderTest` plus three new `ComplaintDAOImplTest` cases: reference-ID match, username/userId fallback match, and literal `%`/`_` wildcard escaping), `complaint.mgt.service`, and the `complaint.mgt.endpoint` webapp (existing tests updated for the new `search` parameter position, plus a new test confirming it's threaded through and that `/me/complaints` always passes `null`). Full suite run as a regression check, all green: `common` 44, `complaint.mgt.dao` 62, `complaint.mgt.service` 153, `complaint.mgt.endpoint` 50 - 309 Java tests total across the touched modules, plus the full frontend suite: **297/297** Vitest tests pass (42 files). JaCoCo's `INSTRUCTION COVEREDRATIO` gate is unaffected on every touched module.

## Integration Tests

**Resolved-complaints fix (#161):** two existing Playwright tests asserted the old hiding behaviour and were updated rather than added to (now `07.05-officer-viewing-complaints.spec.ts` / `07.06-officer-searching-complaints.spec.ts` after this suite's unrelated renumbering on `main`):

- **07.05.03** - retitled from "is hidden from" to "appears in" the default (`status=All`) queue view, with the assertion inverted to `toBeVisible()`. Safe against the shared, never-reset test environment because the default sort is `UPDATED_TIME DESC` (`ComplaintQueryBuilder.java`), so a complaint that was just transitioned lands on page one at 25 rows per page.
- **07.06.03** - its entire premise ("reveals an otherwise-hidden resolved complaint") no longer exists. Rewritten to seed two complaints, one moved to `RESOLVED` and one left `OPEN`, assert both are visible under "All", then assert that selecting "Resolved" keeps the resolved one and drops the open one. This now exercises the server-side status filter rather than a client-side memo.

**Search fix (#201):** `07.06-officer-searching-complaints.spec.ts` already had two tests exercising the search box end-to-end - **07.06.04** ("Searching by reference id narrows the queue to that complaint") and **07.06.05** ("Searching by the Data Principal's name narrows the queue to that principal's complaints"). Neither needed a logic change: they seed a complaint, search for it, and assert every surviving row matches, which is exactly as valid against a real server-side `search` query as it was against the old client-side memo - the fix moved *where* the filtering happens, not what a passing search should show. These are the tests that actually validate #201; against the pre-fix code they'd only have passed by accident (the seeded complaint landing on the loaded page), which is why the file forced `rowsPerPage(25)` for every test up front. That file-level comment, which still described the client-side-only behaviour, has been corrected in this PR to explain that `07.06.04`/`07.06.05` no longer depend on `rowsPerPage` for correctness - it's kept only because `07.06.01`-`03` (status/priority filters, which don't go through `search`) still need it.

**Not executed.** They require a deployed IS plus provisioned personas, and the target deployment's database was cleared during this session. They need the `Action/trigger-e2e` label on the PR.


dpdp-integration-test-suite/tests/07-complaints/05.05-officer-viewing-complaints.spec.ts
```diff
@@ -59,12 +59,11 @@ test.describe('Complaint Officer viewing the queue and a case (UI)', () => {
...
-    // whenever filters.status === 'All' - visible again only once that filter is explicitly set
-    // to "Resolved" (covered by 05.06.03). Asserting on this specific row, not on the word
-    // "Resolved" being absent anywhere on the page - the "Resolved" stat tile's own label makes
-    // that word always present regardless of this filtering behavior.
-    await expect(queuePage.rowByReferenceId(seeded.referenceId)).not.toBeVisible()
+    // "All" means every status, resolved included - ComplaintQueuePage.tsx narrows the
+    // server-paginated page by the reference/name search only, so the pager's total keeps
+    // matching what the table renders. Filtering explicitly to "Resolved" still works and is
+    // covered by 05.06.03.
+    await expect(queuePage.rowByReferenceId(seeded.referenceId)).toBeVisible()
```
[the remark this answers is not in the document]
ParameswaranSajeenthiran: Fixed in https://github.com/wso2/dpdp-accelerator/pull/163/commits/1eab5d2dd531c1355f2f10d428d8280dad3f3e7c

dpdp-accelerator/components/org.wso2.dpdp.accelerator.complaint.mgt.dao/src/main/java/org/wso2/dpdp/accelerator/complaint/mgt/dao/queries/QueryBuilderUtils.java
```diff
@@ -0,0 +1,48 @@
...
+ * KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.wso2.dpdp.accelerator.complaint.mgt.dao.queries;
+
+import java.util.Locale;
+
+// Mirrors org.wso2.dpdp.accelerator.event.notifications.dao.queries.QueryBuilderUtils.
```
hasithakn, line 23: update this comment, lets add a meaning full one

dpdp-accelerator/internal-webapps/org.wso2.dpdp.accelerator.complaint.mgt.endpoint/src/main/resources/complaint-server-API.yaml
```diff
@@ -541,6 +541,14 @@ paths:
           description: Exact match — filter to a specific Data Principal
           schema:
             type: string
+        - name: search
```
hasithakn, line 544: what is this new param added to this swagger?, if this is new shall we have a review?

dpdp-accelerator/components/org.wso2.dpdp.accelerator.complaint.mgt.dao/src/main/java/org/wso2/dpdp/accelerator/complaint/mgt/dao/queries/QueryBuilderUtils.java
ParameswaranSajeenthiran: Updated in https://github.com/wso2/dpdp-accelerator/pull/163/commits/85117a3847e5ec173b09cf65e6b8406efea18c5a

dpdp-accelerator/internal-webapps/org.wso2.dpdp.accelerator.complaint.mgt.endpoint/src/main/resources/complaint-server-API.yaml
ParameswaranSajeenthiran: Added a issue as future improvement - https://github.com/wso2/dpdp-accelerator/issues/222

anjuchamantha (dismissed):

hasithakn (approved):

Issue #159, opened 2026-09-07
Complaint- No usage in having Reply button in the email received to User on receiving the complaints


Ashi1993:
In the first email gets by the user after adding a complaint contains the Review and Reply button, but it doesn't have any usage since nothing to review and reply.

<img width="1437" height="737" alt="Image" src="https://github.com/user-attachments/assets/00353971-f73f-4c06-8dcd-559b0750cf6b" />

ParameswaranSajeenthiran:
this fixes https://github.com/wso2/dpdp-accelerator/pull/164

PR #164, opened 2026-09-07
Remove View & Reply for acknowledgement complaint email

ParameswaranSajeenthiran:
# Purpose

The three complaint notification emails share a single HTML shell (`complaint-email-body.html`), registered for all three template types by `EmailTemplateProvisioningUtil`, with only the subject line differing per type. That shell carried an unconditional **Review & Reply** call-to-action button.

On two of the three types, the button is correct:

* **ComplaintCreated** goes to the Grievance Officer ("filed a new complaint").
* **ComplaintCommentAdded** goes to whichever party did not write the comment ("replied to complaint").

Both recipients genuinely need to act.

On the third, it is wrong. **ComplaintAcknowledged** is sent to the data principal who has just filed the complaint, and `sendAcknowledgement` populates `message-excerpt` with that citizen's own description. The mail therefore quoted the recipient's own words back to them beneath a button instructing them to review and reply — to themselves.

The headline on that type was already correctly passive (`Your complaint {ref} has been received`); the button was the only remaining element implying the recipient owed an action.

## Resolves

Fixes issue - https://github.com/wso2/dpdp-accelerator/issues/159

# Goals

* Remove the **Review & Reply** call-to-action from the complaint acknowledgement mail only.
* Leave the officer ("new complaint filed") and reply ("new reply on complaint") mails byte-identical, since the button is appropriate and load-bearing there.
* Avoid duplicating the shared HTML shell into per-type template files, which would fork the layout and guarantee future drift between the three mails.
* Keep the acknowledgement recipient with a route into the portal, so removing the button does not turn the mail into a dead end.

# Approach

The call-to-action becomes a per-type placeholder, following the pattern the shell already establishes for `{{action-badge-html}}` and `{{headline-html}}` — both of which are HTML fragments computed per notification type in Java rather than hardcoded in the template.

### `complaint-email-body.html`

The seven-line button table is replaced with `{{action-button-html}}`.

The `{{action-url}}` placeholder stays in the file: the footer's quieter **Open in Consent Portal** link uses it, which is what keeps the buttonless acknowledgement navigable.

### `EmailNotificationClient`

* Adds `PLACEHOLDER_ACTION_BUTTON_HTML`.
* Adds a private `buildActionButton(actionUrl)` returning the original markup verbatim.
* `buildTemplatePlaceholders` (created/comment paths) sets it to `buildActionButton(actionUrl)`.
* `sendAcknowledgement` sets it to an empty string.

### Non-obvious constraints

Two non-obvious constraints are documented in comments at the point they apply, because both fail silently and only in a real sent email:

1. **The empty value is set, not omitted.**

   IS's notification handler (`NotificationUtil#extractPlaceHolders`) leaves an unmatched `{{placeholder}}` in the body as literal text rather than dropping it.

   Therefore, omitting the key on the acknowledgement path would ship a visible `{{action-button-html}}` to citizens.

2. **The `href` is interpolated in Java.**

   The URL is not left as `{{action-url}}` inside the button markup.

   Placeholder substitution is a single pass, so a placeholder appearing inside another placeholder's value is never expanded and would render as literal text in the link target.

### UI impact

No portal UI is affected. The change is confined to outbound email bodies, so there is no screenshot to attach.

The rendered difference is the absence of one orange button on one of three mail types.

# User Stories

* As a **Data Principal**, when I file a complaint I receive a confirmation that tells me it was received and does not ask me to act on my own submission.
* As a **Grievance Officer**, I continue to get a direct **Review & Reply** link into the officer view of a newly filed complaint.
* As either party in a complaint thread, I continue to get a direct **Review & Reply** link when the other side replies.

# Release Note

The complaint acknowledgement email sent to a data principal on filing a complaint no longer shows a **Review & Reply** button, which incorrectly asked the recipient to respond to their own complaint.

The officer notification and reply notification emails are unchanged.

# Documentation

**N/A** — no documented behaviour changes.

The email templates are not described field-by-field in the product documentation, and `docs/localization-guide.md` covers localizing admin-created Purposes/Elements, not notification template markup.

The template remains operator-editable via the Console exactly as before.

# Training

**N/A** — no impact on training content.

# Certification

**N/A** — no impact on certification exams.

The change removes a UI element from one notification email; it introduces no new configuration, API, or concept a certification question could target.

# Marketing

**N/A**

# Automation Tests

## Unit Tests

```text
mvn verify -pl dpdp-accelerator/components/org.wso2.dpdp.accelerator.complaint.mgt.service
```

* **153 tests**
* **0 failures**
* **0 errors**
* **0 skipped**
* JaCoCo `INSTRUCTION/COVEREDRATIO` check passed.
* Module coverage: **87.3%** (2841/3255 instructions), against the module's `0.8` minimum.
* `EmailNotificationClient` coverage: **80.7%** (1156/1433 instructions).

No existing test asserted the call-to-action markup, so none needed updating. The four tests referencing `action-badge-html` are unaffected.

Every `{{placeholder}}` in the shell was cross-checked as being populated by both send paths, specifically to prevent the literal-text failure mode described in the approach.

## Integration Tests

None added.

The complaint notification path is not covered by the Playwright suite in `dpdp-integration-test-suite`, which drives the portal UI and has no SMTP capture.

The end-to-end rendering of the acknowledgement email — specifically that the button is gone and no literal `{{action-button-html}}` appears — has not been verified against a live server with a real SMTP sender.

That is the one check worth doing before merge, and it requires a deployed IS with `[output_adapter.email]` configured.

### DIFF
--- a/dpdp-accelerator/components/org.wso2.dpdp.accelerator.complaint.mgt.service/src/main/java/org/wso2/dpdp/accelerator/complaint/mgt/service/notification/EmailNotificationClient.java
+++ b/dpdp-accelerator/components/org.wso2.dpdp.accelerator.complaint.mgt.service/src/main/java/org/wso2/dpdp/accelerator/complaint/mgt/service/notification/EmailNotificationClient.java
@@ -89,6 +89,7 @@ public class EmailNotificationClient implements NotificationClient {
     private static final String PLACEHOLDER_HEADLINE_HTML = "headline-html";
     private static final String PLACEHOLDER_FOOTER_TEXT = "footer-text";
     private static final String PLACEHOLDER_ACTION_BADGE_HTML = "action-badge-html";
+    private static final String PLACEHOLDER_ACTION_BUTTON_HTML = "action-button-html";
     private static final String PLACEHOLDER_LOGO_URL = "logo-url";
 
     private static final String EMAIL_CLAIM = "http://wso2.org/claims/emailaddress";
@@ -273,6 +274,10 @@ public class EmailNotificationClient implements NotificationClient {
                     "You're receiving this because you filed this complaint. We'll email you when there's "
                             + "an update.");
             properties.put(PLACEHOLDER_ACTION_BADGE_HTML, ACKNOWLEDGEMENT_BADGE_HTML);
+            // Deliberately empty, and deliberately still set: IS leaves an unmatched placeholder
+            // in the template as literal text, so omitting the key ships "{{action-button-html}}"
+            // in the mail body. See buildActionButton for why this type has no button.
+            properties.put(PLACEHOLDER_ACTION_BUTTON_HTML, "");
 
             eventService.handleEvent(new Event(IdentityEventConstants.Event.TRIGGER_NOTIFICATION, properties));
         } catch (Throwable t) {
@@ -506,6 +511,7 @@ public class EmailNotificationClient implements NotificationClient {
         placeholders.put(PLACEHOLDER_HEADLINE_HTML, headlineHtml);
         placeholders.put(PLACEHOLDER_FOOTER_TEXT, footerText);
         placeholders.put(PLACEHOLDER_ACTION_BADGE_HTML, actionBadgeHtml);
+        placeholders.put(PLACEHOLDER_ACTION_BUTTON_HTML, buildActionButton(actionUrl));
         placeholders.put(PLACEHOLDER_LOGO_URL, IdentityUtil.getServerURL(LOGO_PATH, true, false));
         return placeholders;
     }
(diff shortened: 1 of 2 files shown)


dpdp-accelerator/components/org.wso2.dpdp.accelerator.complaint.mgt.service/src/main/java/org/wso2/dpdp/accelerator/complaint/mgt/service/notification/EmailNotificationClient.java
```diff
@@ -531,6 +537,20 @@ private static String buildBadge(String text, String background, String color) {
                 + "padding:4px 10px;border-radius:999px;\">" + text + "</span>";
     }
 
+    /**
+     * The URL is interpolated directly since IS's single-pass template substitution won't expand
+     * a placeholder nested inside another placeholder's value.
+     */
+    private static String buildActionButton(String actionUrl) {
+        return "<table role=\"presentation\" cellpadding=\"0\" cellspacing=\"0\" align=\"center\" "
```
hasithakn, line 545: is this ok to define in the code? since we give this templates to customise?

is the base product also follow the same practice?

ParameswaranSajeenthiran:
closing this, since this is overwritten and fixed in the following commit - https://github.com/wso2/dpdp-accelerator/pull/178/commits/98ffec14684332ad4e27c9a2abf342293f99afa6

PR #177, opened 2026-09-10
Add enable/disble Complaint Email config in deployment.toml

### LINKED ISSUE #72: Add Complaints.EmailNotificationsEnabled config switch to opt out of complaint email notifications
Description:

## Why

`EmailNotificationClient` is called synchronously, inline, on the request thread for complaint
creation and commenting — with no config-driven way to skip it. Missing/broken SMTP config can't
fail the complaint submission itself (the DAO write already committed before the client is called,
and it catches `Throwable` everywhere), but "disabled" today just means "hope IS's own
`TRIGGER_NOTIFICATION` handler fails fast silently." An operator without SMTP configured, or who
simply doesn't want this notification channel, has no explicit way to turn it off.

This adds a config switch so that case becomes a deliberate no-op instead of an implicit one, and
removes the latency question entirely when disabled — no `handleEvent()` call happens at all.

## Proposed change

**1. Config chain** (following the repo's standard "adding a new setting" chain):
- `DPDPCommonConstants`: `Complaints.EmailNotificationsEnabled`, default `true` (opt-out, not
  opt-in — existing installs already send email today and shouldn't go silent on upgrade).
- `DPDPConfigParser.isComplaintsEmailNotificationsEnabled()`, same shape as
  `isConsentHistoryEnabled()`.
- Expose via `DPDPConfigurationService`/`DPDPConfigurationServiceImpl`.
- `dpdp-accelerator.xml.j2`: new `<EmailNotificationsEnabled>` element inside `<Complaints>`.
- `wso2is-7.3.0-deployment.toml`: `email_notifications_enabled = true` under
  `[dpdp_accelerator.complaints]`, documented.

**2. The actual switch** — select the `NotificationClient` implementation once, at wiring time,
rather than threading a check into every method:
- New `NoOpNotificationClient implements NotificationClient` — both methods just log-and-return.
- `ComplaintServiceComponent.activate()` picks `EmailNotificationClient` vs
  `NoOpNotificationClient` based on `DPDPConfigurationService.isComplaintsEmailNotificationsEnabled()`.
  `DPDPConfigurationService` is already a mandatory `@Reference` on this component, so it's
  guaranteed bound before `@Activate` runs — no new reference/ordering risk.
- `ComplaintServiceImpl`'s no-arg convenience constructor (test/non-OSGi callers) stays as-is;
  the switch only governs the real production wiring path.

**3. Tests**
- `NoOpNotificationClientTest` — asserts both methods are true no-ops.
- Flag whether `ComplaintServiceComponent` needs its own test now that it has a branch, or stays
  outside the module's tested surface like other `*Component` classes.

**4. Docs** — one line in `docs/configuration-guide.md` alongside the other complaint config knobs.

## Explicitly out of scope

- The synchronous-call **latency** question itself. This switch only helps when the operator
  knows email isn't configured and turns it off — an operator who leaves it `true` with a
  broken/unreachable SMTP host still blocks the request thread on IS's send attempt. Moving the
  notification calls off the request thread (e.g. via an executor) is a separate, larger change.
- Any change to WSO2 IS's own `[output_adapter.email]` / `TRIGGER_NOTIFICATION` handling — stock
  IS behavior, outside this repo's control.

ParameswaranSajeenthiran:
# Purpose

Complaint emails for creation, comment-added, and acknowledgement events are currently sent synchronously and unconditionally by `EmailNotificationClient` whenever SMTP is configured. There is no explicit configuration option to disable this notification channel without breaking SMTP configuration.

# Goals

Add a configuration switch:

```toml
[dpdp_accelerator.complaints]
email_notifications_enabled = false
```

This lets complaint-related email notifications be deliberately enabled through configuration. The flag defaults to `false` — complaint emails are off unless an operator explicitly turns them on.

When disabled, the notification event is not triggered at all, avoiding the `handleEvent()` call and its associated latency.

Fixes  https://github.com/wso2/dpdp-accelerator/issues/72

# Approach

Followed the repository's standard configuration flow for adding a new setting:

`DPDPCommonConstants → DPDPConfigParser → DPDPConfigurationService/Impl → dpdp-accelerator.xml.j2 → deployment.toml`

`EmailNotificationClient`'s two `NotificationClient` methods (`notifyComplaintCreated`, `notifyCommentAdded`) each check `Complaints.EmailNotificationsEnabled` first and return immediately if it's off.

`notifyComplaintCreated`'s check also covers the acknowledgement email, since that is sent from within the same method rather than as a separate `NotificationClient` call.

The configuration service is read through the same test-seam `Supplier` pattern this class already uses for its other OSGi collaborators (`IdentityEventService`, `RealmService`, etc.).

A missing configuration service is treated as enabled, mirroring the null-safe idiom used throughout this codebase for an unresolvable collaborator.

No UI changes.

# User Stories

* As an IS administrator, I want to enable complaint-related emails independently of SMTP configuration, so that the notification channel is on only when I deliberately turn it on rather than being implicit whenever SMTP happens to be configured.
* As an IS administrator, I want complaint emails off by default, so that a fresh install or a staging environment never sends email unless I've explicitly opted in.

# Release Note

Added `[dpdp_accelerator.complaints] email_notifications_enabled` to `deployment.toml`, defaulting to `false`. Complaint-related emails (creation, comments, acknowledgements) must be explicitly enabled, independently of SMTP configuration.

# Documentation

Updated `docs/content/configuration-guide.md`, Section 5 (Configure email notifications), describing the new flag and noting that it is read once at server startup.

# Training

N/A — no training content is affected by this backend configuration flag.

# Certification

N/A — no user-facing behavior or API contract change affects certification exam content.

# Marketing

N/A — this is an internal operational configuration flag, not a customer-facing feature.

# Automation Tests

## Unit Tests

* `EmailNotificationClientTest`: added `notifyComplaintCreatedSendsNothingWhenEmailNotificationsAreDisabled` and `notifyCommentAddedSendsNothingWhenEmailNotificationsAreDisabled`, verifying that no `TRIGGER_NOTIFICATION` event fires when the flag is off.
* `DPDPConfigParserTest`: added `parsesComplaintsEmailNotificationsEnabledValues`, covering key-absent (`→ false`), `"false"`, `"TRUE"`, and an unparseable value (throws `IllegalStateException` rather than silently resolving to `false`).
* Full module test results:

  * `org.wso2.dpdp.accelerator.complaint.mgt.service` — 155 tests passing.
  * `org.wso2.dpdp.accelerator.common` — 42 tests passing.
* JaCoCo `INSTRUCTION/COVEREDRATIO` gate of `0.8` continues to pass for both modules.

## Integration Tests

No automated integration tests were added.

Manually verified end-to-end against a locally deployed WSO2 Identity Server 7.3.0:

* With the flag enabled, the rendered `dpdp-accelerator.xml` contains `EmailNotificationsEnabled=true`.
* With the flag disabled (the default), it contains `EmailNotificationsEnabled=false`.
* The server started successfully in both configurations with no errors.
* `/api/dpdp/complaints/v1/complaints` returned `401` as expected in both configurations.

# Security Checks

* Followed the repository's established configuration-flag conventions.
* FindSecurityBugs was not run.
* Verified via `git diff main...HEAD` that `wso2is-7.3.0-deployment.toml` contains only the new `email_notifications_enabled` key — no credentials.

# Samples

N/A

# Related PRs

None

# Migrations

N/A — no schema or data migration is required. This is a configuration-only addition.

# Test Environment

* **JDK:** 21
* **OS:** macOS (Darwin 25.3.0, arm64)
* **Database:** H2 (embedded `WSO2IDENTITY_DB` and `WSO2DPDP_DB`)
* **Product:** WSO2 Identity Server 7.3.0
* **Browser:** N/A — no UI changes

# Learning

Traced the existing complaint-notification flow — `EmailNotificationClient → IS's native TRIGGER_NOTIFICATION event → IS's notification-template/SMTP pipeline` — to find the cleanest point to apply the check.

Chose a direct check inside `EmailNotificationClient` itself, using the same `Supplier`-based test-seam pattern already established for this class's other collaborators, rather than adding a separate implementation class.


dpdp-accelerator/accelerators/dpdp-is/repository/resources/wso2is-7.3.0-deployment.toml
```diff
@@ -398,6 +398,8 @@ statutory_due_period_days = 90
 attachment_max_size_bytes = 10485760
 # Maximum number of files a single complaint attachment upload request may carry (see AttachmentPolicy).
 attachment_max_files_per_upload = 5
+# Set false to disable complaint-related emails (creation, comments, acknowledgements) entirely
+email_notifications_enabled = true
```
ParameswaranSajeenthiran: make this by default false

dpdp-accelerator/components/org.wso2.dpdp.accelerator.common/src/main/java/org/wso2/dpdp/accelerator/common/constant/DPDPCommonConstants.java
```diff
@@ -53,6 +53,9 @@ public final class DPDPCommonConstants {
     public static final String COMPLAINTS_ATTACHMENT_MAX_FILES_PER_UPLOAD = "Complaints.AttachmentMaxFilesPerUpload";
     public static final int DEFAULT_COMPLAINTS_ATTACHMENT_MAX_FILES_PER_UPLOAD = 5;
 
+    public static final String COMPLAINTS_EMAIL_NOTIFICATIONS_ENABLED = "Complaints.EmailNotificationsEnabled";
+    public static final boolean DEFAULT_COMPLAINTS_EMAIL_NOTIFICATIONS_ENABLED = true;
```
ParameswaranSajeenthiran: make it false by default

dpdp-accelerator/accelerators/dpdp-is/carbon-home/repository/resources/conf/templates/repository/conf/dpdp-accelerator.xml.j2
```diff
@@ -73,6 +73,11 @@
         {% else %}
         <AttachmentMaxFilesPerUpload>5</AttachmentMaxFilesPerUpload>
         {% endif %}
+        {% if dpdp_accelerator.complaints.email_notifications_enabled is defined %}
+        <EmailNotificationsEnabled>{{dpdp_accelerator.complaints.email_notifications_enabled}}</EmailNotificationsEnabled>
+        {% else %}
+        <EmailNotificationsEnabled>true</EmailNotificationsEnabled>
```
ParameswaranSajeenthiran: make it false by default

dpdp-accelerator/components/org.wso2.dpdp.accelerator.complaint.mgt.service/src/main/java/org/wso2/dpdp/accelerator/complaint/mgt/service/notification/EmailNotificationClient.java
```diff
@@ -326,12 +364,12 @@ private List<Recipient> resolveOfficers(String tenantDomain) {
...
             String audience = resolvedAsOrganization ? ROLE_AUDIENCE : "application";
-            if (!roleManagementService.isExistingRoleName(ADMIN_ROLE, audience, organizationId, tenantDomain)) {
-                LOG.debug("Role '" + ADMIN_ROLE + "' does not exist for tenant '" + LogSanitizer.sanitize(tenantDomain)
+            if (!roleManagementService.isExistingRoleName(DPO_ROLE, audience, organizationId, tenantDomain)) {
+                LOG.debug("Role '" + DPO_ROLE + "' does not exist for tenant '" + LogSanitizer.sanitize(tenantDomain)
                         + "'; cannot resolve complaint officers to notify.");
                 return recipients;
             }
-            String roleId = roleManagementService.getRoleIdByName(ADMIN_ROLE, audience, organizationId,
+            String roleId = roleManagementService.getRoleIdByName(DPO_ROLE, audience, organizationId,
```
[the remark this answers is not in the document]
ParameswaranSajeenthiran: The role is provisioned automatically at  tenant creation in src/main/java/org/wso2/dpdp/accelerator/identity/extensions/tenant/DPDPConsentPortalRoleProvisioningUtil.java
        
        RoleV2 dpoRole = createRoleIfNotExists(roleManagementService, DPO_ROLE, toPermissions(dpoScopeNames),
                organizationId, tenantDomain);

dpdp-accelerator/accelerators/dpdp-is/carbon-home/repository/resources/conf/templates/repository/conf/dpdp-accelerator.xml.j2
ParameswaranSajeenthiran: Updated in https://github.com/wso2/dpdp-accelerator/pull/177/changes/f497d2e51d2d7816816af2e12910ca72f01254f5#diff-0aa2540cd5f8520333a33af228f362d9c7583f727713f5404f6e99817facf788

dpdp-accelerator/components/org.wso2.dpdp.accelerator.common/src/main/java/org/wso2/dpdp/accelerator/common/constant/DPDPCommonConstants.java
ParameswaranSajeenthiran: Updated in https://github.com/wso2/dpdp-accelerator/pull/177/changes/f497d2e51d2d7816816af2e12910ca72f01254f5#diff-0aa2540cd5f8520333a33af228f362d9c7583f727713f5404f6e99817facf788

dpdp-accelerator/accelerators/dpdp-is/repository/resources/wso2is-7.3.0-deployment.toml
ParameswaranSajeenthiran: Updated in https://github.com/wso2/dpdp-accelerator/pull/177/changes/f497d2e51d2d7816816af2e12910ca72f01254f5#diff-0aa2540cd5f8520333a33af228f362d9c7583f727713f5404f6e99817facf788

dpdp-accelerator/components/org.wso2.dpdp.accelerator.common/src/main/java/org/wso2/dpdp/accelerator/common/constant/DPDPCommonConstants.java
```diff
@@ -53,6 +53,9 @@ public final class DPDPCommonConstants {
     public static final String COMPLAINTS_ATTACHMENT_MAX_FILES_PER_UPLOAD = "Complaints.AttachmentMaxFilesPerUpload";
     public static final int DEFAULT_COMPLAINTS_ATTACHMENT_MAX_FILES_PER_UPLOAD = 5;
 
+    public static final String COMPLAINTS_EMAIL_NOTIFICATIONS_ENABLED = "Complaints.EmailNotificationsEnabled";
+    public static final boolean DEFAULT_COMPLAINTS_EMAIL_NOTIFICATIONS_ENABLED = false;
```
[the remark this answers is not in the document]
ParameswaranSajeenthiran: Updated the PR description

dpdp-accelerator/components/org.wso2.dpdp.accelerator.complaint.mgt.service/src/main/java/org/wso2/dpdp/accelerator/complaint/mgt/service/notification/EmailNotificationClient.java
```diff
@@ -100,10 +101,12 @@ public class EmailNotificationClient implements NotificationClient {
...
     // Mirrors identity.extensions' DPDPConsentPortalAppProvisioningUtil.APPLICATION_NAME /
-    // DPDPConsentPortalRoleProvisioningUtil.ADMIN_ROLE - duplicated rather than depended on, same
-    // as every other constant in this class.
+    // DPDPConsentPortalRoleProvisioningUtil.DPO_ROLE - duplicated rather than depended on, same
+    // as every other constant in this class. dpdp-consent-dpo (not dpdp-consent-admin) is the
+    // complaint officer distribution list - it already carries complaints:read:any/write:any,
+    // so its members can act on what they're notified about.
     private static final String APPLICATION_NAME = "DPDP Consent Portal";
-    private static final String ADMIN_ROLE = "dpdp-consent-admin";
+    private static final String DPO_ROLE = "dpdp-consent-dpo";
```
hasithakn: in DPDPConsentPortalRoleProvisioningUtil we have all roles defined in strings, as below

```
    public static final String ADMIN_ROLE = "dpdp-consent-admin";
    static final String USER_ROLE = "dpdp-consent-user";
    public static final String DPO_ROLE = "dpdp-consent-dpo";
```

shall we move these to common modules, `DPDPCommonConstants.java` , and refer in this class?

dpdp-accelerator/components/org.wso2.dpdp.accelerator.complaint.mgt.service/src/main/java/org/wso2/dpdp/accelerator/complaint/mgt/service/notification/EmailNotificationClient.java
```diff
@@ -326,12 +364,12 @@ private List<Recipient> resolveOfficers(String tenantDomain) {
...
-                LOG.debug("Role '" + ADMIN_ROLE + "' does not exist for tenant '" + LogSanitizer.sanitize(tenantDomain)
+            if (!roleManagementService.isExistingRoleName(DPO_ROLE, audience, organizationId, tenantDomain)) {
+                LOG.debug("Role '" + DPO_ROLE + "' does not exist for tenant '" + LogSanitizer.sanitize(tenantDomain)
                         + "'; cannot resolve complaint officers to notify.");
                 return recipients;
             }
-            String roleId = roleManagementService.getRoleIdByName(ADMIN_ROLE, audience, organizationId,
+            String roleId = roleManagementService.getRoleIdByName(DPO_ROLE, audience, organizationId,
                     tenantDomain);
             List<UserBasicInfo> members = roleManagementService.getUserListOfRole(roleId, tenantDomain);
```
hasithakn, line 376: lets add a null or is empty check for roleId right?

dpdp-accelerator/components/org.wso2.dpdp.accelerator.complaint.mgt.service/src/test/java/org/wso2/dpdp/accelerator/complaint/mgt/service/notification/EmailNotificationClientTest.java
```diff
@@ -68,7 +69,7 @@ public class EmailNotificationClientTest {
 
     private static final String EMAIL_CLAIM = "http://wso2.org/claims/emailaddress";
     private static final String APPLICATION_NAME = "DPDP Consent Portal";
-    private static final String ADMIN_ROLE = "dpdp-consent-admin";
+    private static final String DPO_ROLE = "dpdp-consent-dpo";
```
hasithakn: as i understand you can re use the common modules variable, rather than re redefining right?

hasithakn:
lets remove package-lock json as no UI is changed from this PR

hasithakn:
One design question on `EmailNotificationClient.emailNotificationsEnabled()`:

```java
private boolean emailNotificationsEnabled() {
    DPDPConfigurationService configurationService = configurationServiceSupplier.get();
    return configurationService == null || configurationService.isComplaintsEmailNotificationsEnabled();
}
```

This fails **open**: if the `DPDPConfigurationService` OSGi reference can't be resolved, emails are treated as *enabled*. That's the opposite of this feature's own stated default ("off unless an operator explicitly turns it on"). `DPDPConfigurationServiceImpl` itself takes the opposite stance for the equivalent failure (missing `configParser` → falls back to `DEFAULT_COMPLAINTS_EMAIL_NOTIFICATIONS_ENABLED`, i.e. disabled), so there's an inconsistency between this class and its own service layer for the same kind of failure.

I get the rationale in the comment (mirroring how other collaborators like `IdentityEventService`/`RealmService` fail open) — but those are core IS services where a null is a genuine wiring bug; `DPDPConfigurationService` is this accelerator's own service, and fail-open here means a rare OSGi resolution hiccup could silently send emails an operator never opted into, which undercuts the whole point of this PR.

Suggested fix — flip to fail-closed:

```java
private boolean emailNotificationsEnabled() {
    DPDPConfigurationService configurationService = configurationServiceSupplier.get();
    return configurationService != null && configurationService.isComplaintsEmailNotificationsEnabled();
}
```

One test will need a small update after this change: `notifyComplaintCreatedSendsOnlyTheAcknowledgementWhenNoOfficersResolve` uses the 4-arg constructor, which defaults `configurationServiceSupplier` to `() -> null` — today that means "enabled" (fail-open), so its `verify(identityEventService, times(1)).handleEvent(...)` currently passes only because of the bug above. After the fix it'd need the 6-arg constructor with a mocked `configurationService` stubbed to return `true`, since this test's actual intent is "acknowledgement still sends with no officers," not exercising the toggle itself. The other two tests using the same short constructor (`notifyComplaintCreatedSendsNothingWhenRealmServiceIsUnresolvable`, `acknowledgementIsSkippedWhenTheEmailClaimIsUnresolvable`) already assert `never()`, so they're unaffected.

Separately (not a blocker, just flagging): this PR also silently changes who gets notified — `resolveOfficers()` now checks `dpdp-consent-dpo` membership instead of `dpdp-consent-admin`. Looks intentional and is documented well (code comments, tests, and `configuration-guide.md` all updated consistently), but worth calling out explicitly in the PR description for anyone upgrading an existing deployment — a tenant with `dpdp-consent-admin` members but nobody yet in `dpdp-consent-dpo` would see zero complaint officers notified once this flag is turned on.

ParameswaranSajeenthiran:
> Implement review suggestions



> One design question on `EmailNotificationClient.emailNotificationsEnabled()`:
> 
> ```java
> private boolean emailNotificationsEnabled() {
>     DPDPConfigurationService configurationService = configurationServiceSupplier.get();
>     return configurationService == null || configurationService.isComplaintsEmailNotificationsEnabled();
> }
> ```
> 
> This fails **open**: if the `DPDPConfigurationService` OSGi reference can't be resolved, emails are treated as _enabled_. That's the opposite of this feature's own stated default ("off unless an operator explicitly turns it on"). `DPDPConfigurationServiceImpl` itself takes the opposite stance for the equivalent failure (missing `configParser` → falls back to `DEFAULT_COMPLAINTS_EMAIL_NOTIFICATIONS_ENABLED`, i.e. disabled), so there's an inconsistency between this class and its own service layer for the same kind of failure.
> 
> I get the rationale in the comment (mirroring how other collaborators like `IdentityEventService`/`RealmService` fail open) — but those are core IS services where a null is a genuine wiring bug; `DPDPConfigurationService` is this accelerator's own service, and fail-open here means a rare OSGi resolution hiccup could silently send emails an operator never opted into, which undercuts the whole point of this PR.
> 
> Suggested fix — flip to fail-closed:
> 
> ```java
> private boolean emailNotificationsEnabled() {
>     DPDPConfigurationService configurationService = configurationServiceSupplier.get();
>     return configurationService != null && configurationService.isComplaintsEmailNotificationsEnabled();
> }
> ```
> 
> One test will need a small update after this change: `notifyComplaintCreatedSendsOnlyTheAcknowledgementWhenNoOfficersResolve` uses the 4-arg constructor, which defaults `configurationServiceSupplier` to `() -> null` — today that means "enabled" (fail-open), so its `verify(identityEventService, times(1)).handleEvent(...)` currently passes only because of the bug above. After the fix it'd need the 6-arg constructor with a mocked `configurationService` stubbed to return `true`, since this test's actual intent is "acknowledgement still sends with no officers," not exercising the toggle itself. The other two tests using the same short constructor (`notifyComplaintCreatedSendsNothingWhenRealmServiceIsUnresolvable`, `acknowledgementIsSkippedWhenTheEmailClaimIsUnresolvable`) already assert `never()`, so they're unaffected.
> 
> Separately (not a blocker, just flagging): this PR also silently changes who gets notified — `resolveOfficers()` now checks `dpdp-consent-dpo` membership instead of `dpdp-consent-admin`. Looks intentional and is documented well (code comments, tests, and `configuration-guide.md` all updated consistently), but worth calling out explicitly in the PR description for anyone upgrading an existing deployment — a tenant with `dpdp-consent-admin` members but nobody yet in `dpdp-consent-dpo` would see zero complaint officers notified once this flag is turned on.

implemented in the following commit - 
https://github.com/wso2/dpdp-accelerator/pull/177/commits/743ce0d8d2e2626c66692c20d4fa3e8d1db31de3

dpdp-accelerator/components/org.wso2.dpdp.accelerator.complaint.mgt.service/src/test/java/org/wso2/dpdp/accelerator/complaint/mgt/service/notification/EmailNotificationClientTest.java
ParameswaranSajeenthiran: Updated in https://github.com/wso2/dpdp-accelerator/pull/177/commits/743ce0d8d2e2626c66692c20d4fa3e8d1db31de3

dpdp-accelerator/components/org.wso2.dpdp.accelerator.complaint.mgt.service/src/main/java/org/wso2/dpdp/accelerator/complaint/mgt/service/notification/EmailNotificationClient.java
ParameswaranSajeenthiran: Udpated in https://github.com/wso2/dpdp-accelerator/pull/177/commits/743ce0d8d2e2626c66692c20d4fa3e8d1db31de3

dpdp-accelerator/components/org.wso2.dpdp.accelerator.complaint.mgt.service/src/main/java/org/wso2/dpdp/accelerator/complaint/mgt/service/notification/EmailNotificationClient.java
ParameswaranSajeenthiran: Updated in https://github.com/wso2/dpdp-accelerator/commit/743ce0d8d2e2626c66692c20d4fa3e8d1db31de3

docs/content/configuration-guide.md
```diff
@@ -1,8 +1,3 @@
----
-title: Configuring the Consent Portal application
-sidebar_position: 2
----
-
```
hasithakn, line 5: isnt this change going to remove this page form doc page?

https://wso2.github.io/dpdp-accelerator/docs/configuration-guide

dpdp-accelerator/internal-webapps/org.wso2.dpdp.accelerator.complaint.mgt.endpoint/pom.xml
```diff
@@ -57,7 +57,7 @@
 
         <!--
             complaint.mgt.service's NotificationClient references IdentityEventService, (for the
-            officer deep link) ApplicationManagementService, and (for resolving dpdp-consent-admin
+            officer deep link) ApplicationManagementService, and (for resolving dpdp-consent-dpo
```
hasithakn, line 60: Shall we remove package-lock.json‎ changes? since you have not done UI changes here?

docs/content/configuration-guide.md
ParameswaranSajeenthiran: Added back in https://github.com/wso2/dpdp-accelerator/pull/177/commits/2f2d2e8e599118a417f86739114cfdc11f121419

dpdp-accelerator/internal-webapps/org.wso2.dpdp.accelerator.complaint.mgt.endpoint/pom.xml
ParameswaranSajeenthiran: Reverted the change in https://github.com/wso2/dpdp-accelerator/pull/177/commits/2f2d2e8e599118a417f86739114cfdc11f121419

hasithakn (approved):

Issue #73, opened 2026-09-01
Make complaint email templates customizable at server level without rebuilding


hasithakn:
Description:

## Current state

Complaint email templates are provisioned into WSO2 IS's own native, per-tenant email template
store (`NotificationTemplateManager`, the same mechanism behind Console's "Email Templates" UI) —
`EmailTemplateProvisioningUtil.provisionTemplates()`
(`components/org.wso2.dpdp.accelerator.identity.extensions/.../notification/EmailTemplateProvisioningUtil.java`)
calls IS's own `addNotificationTemplateType()`/`addNotificationTemplate()` for three template
types: `ComplaintCreated`, `ComplaintCommentAdded`, `ComplaintAcknowledged`. Delivery goes through
IS's own `TRIGGER_NOTIFICATION` handler and `[output_adapter.email]` SMTP config — nothing here is
bypassed or reinvented.

So in principle, an admin can already open Console → Email Templates and edit these three
templates per tenant, today, with no rebuild. **In practice this doesn't work as a real
customization path**, for two concrete reasons:

1. **Every edit gets silently overwritten.** `DPDPIdentityExtensionTenantMgtListener` calls
   `provisionTemplates(tenantDomain)` unconditionally from both `onTenantCreate` *and*
   `onTenantUpdate` — not just once at first provisioning. `provisionTemplate()`'s own javadoc says
   it "always (re)writes the template content," and the existing test suite
   (`EmailTemplateProvisioningUtilTest`) explicitly documents this as the intended upgrade path
   ("re-running this on every tenant startup is exactly how an updated template body reaches an
   already-provisioned tenant"). The consequence: any tenant-update event (not even an accelerator
   upgrade — any tenant metadata change) resets a customized template straight back to the bundled
   default, with no warning to the admin who edited it.
2. **The subject lines and the shared HTML body are themselves 100% build-time content** — the
   three subjects are Java string literals in `EmailTemplateProvisioningUtil` (lines 59-64), and
   all three types share one bundled classpath resource,
   `identity.extensions/src/main/resources/notification/complaint-email-body.html`. There's no
   documentation anywhere (`docs/configuration-guide.md`, `docs/localization-guide.md`) telling an
   operator that Console's Email Templates screen is even the relevant place to look — the
   localization guide only covers Purposes/Elements/i18n wording, not email templates at all.

Net effect: changing wording, branding, or copy in a complaint notification email today requires
[...]

PR #178, opened 2026-09-10
Preserve customized complaint email templates

ParameswaranSajeenthiran:
# Purpose

Complaint email templates (`ComplaintCreated`, `ComplaintCommentAdded`, `ComplaintAcknowledged`) are provisioned into WSO2 IS's own per-tenant notification template store — the same store edited by the Console's **Email Templates** screen.

In practice, this wasn't usable as a real customization path because `DPDPIdentityExtensionTenantMgtListener` called `provisionTemplates()` unconditionally on every `onTenantCreate` / `onTenantUpdate`. As a result, any tenant-update event silently reset an administrator's Console edit back to the bundled default.

There was also no documented way to change the bundled default itself without a full Java rebuild, since the HTML body and subjects were hardcoded as a classpath resource and Java string literals.

**Fixes 
- https://github.com/wso2/dpdp-accelerator/issues/73  
- https://github.com/wso2/dpdp-accelerator/issues/159

## Goals

* An administrator's Console edit to a complaint template survives every future tenant-update event and server restart permanently.
* The bundled default subject/body itself can be changed by an operator without a Java rebuild.
* Document the customization path and the full placeholder set for operators.

## Approach

* `EmailTemplateProvisioningUtil.provisionTemplate()` now performs a permanent check-then-add per tenant. This mirrors the "provisioning is idempotent, never assume a clean slate" convention already used elsewhere in this listener.
* A template is written once, the first time a tenant is provisioned, and never touched again. There is no flag or action that overwrites an existing template.
* A template-existence lookup failure (any error other than the manager's own "not found" code) is treated as "skip, don't write" rather than "not present" — a real lookup failure could be hiding an existing Console customization, and provisioning retries safely on the next tenant-update event.
* Added `EmailTemplateConfigLoader`, a new class that reads the bundled default subject/body from `<IS_HOME>/repository/conf/email/email-dpdp-config.xml` when present.
* This mirrors the exact shape and lifecycle of the product's own `email-admin-config.xml`. This was verified against the actual `org.wso2.carbon.email.mgt` bundle: IS's built-in templates work the same way — a bundled file seeds an in-memory-only fallback, consulted only when no DB row exists yet for that tenant.
* The config file is parsed defensively: DTD and external-entity resolution are disabled on the StAX reader (XXE hardening), and any parse failure — or the file being entirely absent — falls back to the classpath default rather than blocking provisioning.
* Falls back to the existing classpath resource (`complaint-email-body.html`) if the file is missing, unparseable, or doesn't define a given type.
* No UI changes — backend/configuration only.

## User Stories

* As an operator, I can customize a complaint notification email's subject/body in Console and trust that it stays customized indefinitely, instead of silently reverting on the next tenant update.
* As an operator preparing a deployment, I can change what new tenants get by default by editing `email-dpdp-config.xml`, with no rebuild.

## Release Note

> **Fixed:** Complaint notification email templates (Console → Email Templates) are no longer silently overwritten on tenant updates — a Console edit is now permanent.
>
> Added `email-dpdp-config.xml` as an operator-editable source for the bundled default template content.

## Documentation

* New page: `docs/content/complaint-management-guide.md` ("Configuring Complaint Management"), registered in `docs/sidebars.js` right after `event-notification-guide`.

  * "Customizing complaint notification emails" section
  * Where to edit templates in Console
  * Full placeholder reference (`{{reference-id}}`, `{{message-excerpt}}`, `{{data-principal-name}}`, `{{actor-name}}`, `{{category-label}}`, `{{priority-label}}`, `{{status-label}}`, `{{sla-label}}`, `{{action-url}}`, `{{recipient-role-label}}`, `{{headline-html}}`, `{{footer-text}}`, `{{action-badge-html}}`, `{{logo-url}}`)
  * Permanent check-then-add behavior
  * `email-dpdp-config.xml` configuration file — including that it's read once and cached for the server's lifetime, so a server restart is required after editing it before new tenants pick up the change

## Training

**N/A** — No training content impact.

## Certification

**N/A** — Internal provisioning-idempotency fix and a new operator configuration file; no user-facing product behavior change warranting new certification questions.

## Marketing

**N/A** — Internal bug fix, not a marketed feature.

## Automation Tests

* **Unit tests**

  * `EmailTemplateProvisioningUtilTest` (updated, 6 tests)

    * First-time provisioning writes all three templates
    * An already-registered type not blocking content write
    * Write failures not throwing
    * An existing template being left untouched
    * Content being written when the template is genuinely not found yet
    * An existence-check failure being treated as "skip, don't overwrite"
  * `EmailTemplateConfigLoaderTest` (new, 8 tests)

    * No config directory resolvable
    * File missing
    * Valid file read
    * Type not present in file
    * An incomplete `<configuration>` element
    * Malformed XML
    * A zero-byte file
    * External-entity resolution is blocked (XXE hardening)
  * `mvn clean verify` passes on `common` and `identity.extensions`, with JaCoCo coverage gates satisfied.

* **Integration tests**

  * None added to `dpdp-integration-test-suite`.
  * Verified manually end-to-end instead, including reading IS's own `org.wso2.carbon.email.mgt` bundle to confirm the design matches the product's existing precedent.

## Security Checks

* Followed secure coding standards in [WSO2 Secure Engineering Guidelines](http://wso2.com/technical-reports/wso2-secure-engineering-guidelines): **Yes**
* Ran the FindSecurityBugs plugin and verified the report: **No issues**
* Confirmed that this PR doesn't commit any keys, passwords, tokens, usernames, or other secrets: **Yes**
* `EmailTemplateConfigLoader`'s XML parsing disables DTD and external-entity resolution, closing an XXE avenue on a file that (per the design) an operator, not an attacker, is expected to control — hardened anyway since it's parsed on every cold cache miss.

## Samples

**N/A**

## Related PRs

**None**

## Migrations

**N/A** — No database schema change.

## Test Environment

* **JDK:** 21 (Homebrew OpenJDK)
* **OS:** macOS (aarch64)
* **Databases:**

  * H2 (embedded)
  * MySQL 8.4.11 (`is-mysql-deploy`, port `3307`)

### Verification

* Built the full accelerator (`mvn clean install` from the repository root) and deployed it over a MySQL-backed IS instance via `bin/merge.sh`.
* Edited a template via the Console REST API.
* Restarted the server and confirmed the edit survived:

  * Via the API
  * Directly against `IDN_NOTIFICATION_ORG_TEMPLATE` for MySQL
* Confirmed that a brand-new tenant is unaffected by an existing tenant's edit.
* Confirmed that editing `email-dpdp-config.xml` only affects tenants provisioned after the edit.

## Learning

Decompiled and read the actual `org.wso2.carbon.email.mgt` classes:

* `UnifiedTemplateManager`
* `SystemDefaultTemplateManager`
* `DBBasedTemplateManager`

This confirmed that IS's own notification-template architecture already follows the exact **"DB row wins forever, bundled file is only a one-time seed"** pattern.

This established that:

1. The unconditional-overwrite code was a bug specific to the accelerator and was not inherited from IS.
2. The design for `email-dpdp-config.xml` is consistent with IS's existing architecture.
3. The `getNotificationTemplate` `null`-means-not-found assumption is valid.

### DIFF
--- a/docs/content/configuration-guide.md
+++ b/docs/content/configuration-guide.md
@@ -368,12 +368,20 @@ Available placeholders: `{{reference-id}}`, `{{message-excerpt}}`,
 `{{action-badge-html}}`, `{{logo-url}}` (see `EmailNotificationClient` for
 exactly what each resolves to).
 
-Provisioning is check-then-add per tenant, so a Console edit is never
-overwritten by a later tenant update. To force every tenant's templates back
-to the bundled default (e.g. after an accelerator upgrade), set
-`reset_to_default_enabled = true` under `[dpdp_accelerator.complaints.email_templates]`,
-trigger a tenant update, then set it back to `false` - otherwise every future
-update keeps re-clobbering the templates.
+Provisioning is check-then-add per tenant: a template is written once, the
+first time a tenant is provisioned, and never touched again after that - a
+Console edit is permanent and survives every later tenant-update event.
+There is no flag or action that resets an already-provisioned tenant's
+template back to the bundled default; the only way to change what a tenant
+already has is to edit it again in Console. A change to the bundled default
+(see below) only affects tenants provisioned after that change.
+
+The bundled default subject/body itself comes from
+`<IS_HOME>/repository/conf/email/email-dpdp-config.xml` when present, the
+same shape as this product's own `email-admin-config.xml`. Edit that file to
+change the default new tenants get, with no Java rebuild - falls back to the
+accelerator's own bundled resource if the file is missing or doesn't define a
+given type.
 
 All three types currently share one bundled HTML body; splitting them per-type
 is a possible future improvement, not yet decided.
--- a/docs/content/configuration-guide.md
+++ b/docs/content/configuration-guide.md
@@ -352,3 +352,28 @@ pending_subscription_recovery_threshold_seconds = 60
 
 These are server-wide runtime settings. Subscription `shared_secret` values
 remain per-subscription data and are not placed in `dpdp-accelerator.xml`.
+
+# Customizing complaint email templates
+
+The three complaint notification emails (`ComplaintCreated`,
+`ComplaintCommentAdded`, `ComplaintAcknowledged`) are standard IS notification
+templates - edit their subject/body in Console under **Email Templates**, per
+tenant, the same as any other IS template. No rebuild needed; delivery goes
+through the usual `[output_adapter.email]` SMTP config.
+
+Available placeholders: `{{reference-id}}`, `{{message-excerpt}}`,
+`{{data-principal-name}}`, `{{actor-name}}`, `{{category-label}}`,
+`{{priority-label}}`, `{{status-label}}`, `{{sla-label}}`, `{{action-url}}`,
+`{{recipient-role-label}}`, `{{headline-html}}`, `{{footer-text}}`,
+`{{action-badge-html}}`, `{{logo-url}}` (see `EmailNotificationClient` for
+exactly what each resolves to).
+
+Provisioning is check-then-add per tenant, so a Console edit is never
+overwritten by a later tenant update. To force every tenant's templates back
+to the bundled default (e.g. after an accelerator upgrade), set
+`reset_to_default_enabled = true` under `[dpdp_accelerator.complaints.email_templates]`,
+trigger a tenant update, then set it back to `false` - otherwise every future
+update keeps re-clobbering the templates.
+
+All three types currently share one bundled HTML body; splitting them per-type
+is a possible future improvement, not yet decided.
--- a/dpdp-accelerator/accelerators/dpdp-is/repository/resources/wso2is-7.3.0-deployment.toml
+++ b/dpdp-accelerator/accelerators/dpdp-is/repository/resources/wso2is-7.3.0-deployment.toml
@@ -399,6 +399,16 @@ attachment_max_size_bytes = 10485760
 # Maximum number of files a single complaint attachment upload request may carry (see AttachmentPolicy).
 attachment_max_files_per_upload = 5
 
+# The three complaint notification email templates (ComplaintCreated, ComplaintCommentAdded,
+# ComplaintAcknowledged) are provisioned once per tenant into Console -> Email Templates and left
+# alone after that, so an administrator's edit there survives future tenant-update events. Set
+# reset_to_default_enabled = true to instead force every tenant's templates back to the bundled
+# subject/body on the next tenant-update event (e.g. right after upgrading the accelerator) -
+# remember to set it back to false afterwards, or every future update keeps re-clobbering them.
+# See docs/configuration-guide.md's "Customizing complaint email templates" section.
+[dpdp_accelerator.complaints.email_templates]
+reset_to_default_enabled = false
+
 # Event Notification Framework runtime settings.
 [dpdp_accelerator.event_notifications]
 # Automatically reconciles the DPDP system topics (consent.update, consent.revoke, etc.) for
--- a/dpdp-accelerator/components/org.wso2.dpdp.accelerator.identity.extensions/pom.xml
+++ b/dpdp-accelerator/components/org.wso2.dpdp.accelerator.identity.extensions/pom.xml
@@ -54,6 +54,10 @@
             <artifactId>org.wso2.carbon.utils</artifactId>
         </dependency>
         <dependency>
+            <groupId>org.apache.ws.commons.axiom.wso2</groupId>
+            <artifactId>axiom</artifactId>
+        </dependency>
+        <dependency>
             <groupId>org.wso2.carbon</groupId>
             <artifactId>org.wso2.carbon.user.core</artifactId>
         </dependency>
@@ -242,8 +246,11 @@
                             org.apache.commons.logging;version="${commons.logging.imp.pkg.version.range}",
                             com.google.gson;version="${gson.imp.pkg.version.range}",
 
+                            org.apache.axiom.*;version="0",
+
                             org.wso2.carbon.context.*;version="${carbon.kernel.version}",
                             org.wso2.carbon.core;version="${carbon.kernel.version}",
+                            org.wso2.carbon.utils;version="${carbon.kernel.version}",
                             org.wso2.carbon.utils.multitenancy;version="${carbon.kernel.version}",
                             org.wso2.carbon.user.core;version="${carbon.user.core.version}",
                             org.wso2.carbon.user.core.common;version="${carbon.user.core.version}",
--- a/dpdp-accelerator/components/org.wso2.dpdp.accelerator.identity.extensions/src/main/java/org/wso2/dpdp/accelerator/identity/extensions/notification/EmailTemplateConfigLoader.java
+++ b/dpdp-accelerator/components/org.wso2.dpdp.accelerator.identity.extensions/src/main/java/org/wso2/dpdp/accelerator/identity/extensions/notification/EmailTemplateConfigLoader.java
@@ -1,0 +1,171 @@
+/*
+ * Copyright (c) 2026, WSO2 LLC. (https://www.wso2.com).
+ *
+ * WSO2 LLC. licenses this file to you under the Apache License,
+ * Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.wso2.dpdp.accelerator.identity.extensions.notification;
+
+import org.apache.axiom.om.OMElement;
+import org.apache.axiom.om.OMException;
+import org.apache.axiom.om.impl.builder.StAXOMBuilder;
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+import org.wso2.carbon.utils.CarbonUtils;
+
+import java.io.File;
+import java.io.IOException;
+import java.io.InputStream;
+import java.nio.file.Files;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Iterator;
+import java.util.Map;
+import java.util.Optional;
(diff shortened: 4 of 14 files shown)


docs/content/configuration-guide.md
```diff
@@ -352,3 +352,36 @@ pending_subscription_recovery_threshold_seconds = 60
...
 These are server-wide runtime settings. Subscription `shared_secret` values
 remain per-subscription data and are not placed in `dpdp-accelerator.xml`.
+
+# Customizing complaint email templates
+
+The three complaint notification emails (`ComplaintCreated`,
+`ComplaintCommentAdded`, `ComplaintAcknowledged`) are standard IS notification
+templates - edit their subject/body in Console under **Email Templates**, per
+tenant, the same as any other IS template. No rebuild needed; delivery goes
+through the usual `[output_adapter.email]` SMTP config.
```
ParameswaranSajeenthiran: Remove this

docs/content/configuration-guide.md
```diff
@@ -352,3 +352,36 @@ pending_subscription_recovery_threshold_seconds = 60
 
 These are server-wide runtime settings. Subscription `shared_secret` values
 remain per-subscription data and are not placed in `dpdp-accelerator.xml`.
+
+# Customizing complaint email templates
```
ParameswaranSajeenthiran: Move this to complaint docs

dpdp-accelerator/accelerators/dpdp-is/repository/resources/wso2is-7.3.0-deployment.toml
```diff
@@ -399,6 +399,16 @@ attachment_max_size_bytes = 10485760
 # Maximum number of files a single complaint attachment upload request may carry (see AttachmentPolicy).
 attachment_max_files_per_upload = 5
 
+# The three complaint notification email templates (ComplaintCreated, ComplaintCommentAdded,
```
ParameswaranSajeenthiran: summarize

ParameswaranSajeenthiran: review this, do we really need this

dpdp-accelerator/components/org.wso2.dpdp.accelerator.identity.extensions/src/main/java/org/wso2/dpdp/accelerator/identity/extensions/notification/EmailTemplateConfigLoader.java
```diff
@@ -0,0 +1,171 @@
...
+
+        synchronized (LOCK) {
+            if (templatesByType == null) {
+                templatesByType = loadTemplates();
+            }
+            return templatesByType;
+        }
+    }
+
+    private static Map<String, TemplateContent> loadTemplates() {
```
ParameswaranSajeenthiran, line 85: check how IS does this and reuse the methods

docs/content/configuration-guide.md
ParameswaranSajeenthiran: Removed in https://github.com/wso2/dpdp-accelerator/pull/178/commits/0071e6a249365acd6f4028f7e25e5c1a5487e54b

dpdp-accelerator/components/org.wso2.dpdp.accelerator.identity.extensions/src/main/java/org/wso2/dpdp/accelerator/identity/extensions/notification/EmailTemplateConfigLoader.java
ParameswaranSajeenthiran: I checked what's actually in the Maven dependency graph (org.wso2.carbon.identity.governance:1.8.107, the only artifact this module depends on for NotificationTemplateManager):   
                                                                                                                                                                                     
  - NotificationTemplateManager is just an interface with no-op default methods. The concrete implementation that parses IS's real repository/conf/email/email-admin-config.xml at   
    server startup lives inside the actual IS product distribution — it's not published as a reusable Maven artifact anywhere in ~/.m2, and isn't something an OSGi extension bundle 
    like ours can call into directly.                                                                                                                                                
  - The interface does expose getDefaultNotificationTemplates(channel) / addDefaultNotificationTemplates(channel, tenantDomain) — but that's IS bulk-seeding its own built-in        
    template types (account confirmation, password reset, etc.) that it already knows about internally. There's no public extension point for a third-party bundle to register a     
    brand-new template type (ComplaintCreated, etc.) into that same default-template catalog. So those methods can't help us here.                                                   
                                                                                                                                                                                     
  So the only real "IS way" available to us is exactly what EmailTemplateProvisioningUtil already does: build a NotificationTemplate object and call                                 
  templateManager.addNotificationTemplate(...). There's no shared loader to plug our own config file into upstream of that call — IS doesn't expose one for custom types.

docs/content/configuration-guide.md
```diff
@@ -352,3 +352,36 @@ pending_subscription_recovery_threshold_seconds = 60
...
+`{{data-principal-name}}`, `{{actor-name}}`, `{{category-label}}`,
+`{{priority-label}}`, `{{status-label}}`, `{{sla-label}}`, `{{action-url}}`,
+`{{recipient-role-label}}`, `{{headline-html}}`, `{{footer-text}}`,
+`{{action-badge-html}}`, `{{logo-url}}` (see `EmailNotificationClient` for
+exactly what each resolves to).
+
+Provisioning is check-then-add per tenant: a template is written once, the
+first time a tenant is provisioned, and never touched again after that - a
+Console edit is permanent and survives every later tenant-update event.
+There is no flag or action that resets an already-provisioned tenant's
```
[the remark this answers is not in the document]
ParameswaranSajeenthiran: removed in https://github.com/wso2/dpdp-accelerator/pull/178/commits/dc4f0c057eec6ae0ab532ccb5bda691b4d6016c5

dpdp-accelerator/accelerators/dpdp-is/repository/resources/wso2is-7.3.0-deployment.toml
ParameswaranSajeenthiran: Remove in https://github.com/wso2/dpdp-accelerator/pull/178/commits/dc4f0c057eec6ae0ab532ccb5bda691b4d6016c5

docs/content/configuration-guide.md
ParameswaranSajeenthiran: Moved in https://github.com/wso2/dpdp-accelerator/pull/178/commits/3a63a4547eef25f7594faee182b366f925d0dd53

dpdp-accelerator/components/org.wso2.dpdp.accelerator.identity.extensions/src/main/java/org/wso2/dpdp/accelerator/identity/extensions/notification/EmailTemplateProvisioningUtil.java
```diff
@@ -82,22 +92,38 @@ private static String loadResource(String path) {
...
             LOG.debug("Notification template type '" + templateType + "' already registered for tenant '"
-                    + LogSanitizer.sanitize(tenantDomain) + "'; continuing to (re)write its content.", e);
+                    + LogSanitizer.sanitize(tenantDomain) + "'; continuing to check its content.", e);
+        }
+
+        if (templateExists(templateManager, templateType, tenantDomain)) {
+            LOG.debug("Email template '" + templateType + "' already exists for tenant '"
+                    + LogSanitizer.sanitize(tenantDomain) + "'; leaving its content as-is so a Console "
+                    + "customization is preserved.");
+            return;
```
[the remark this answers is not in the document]
ParameswaranSajeenthiran: removed in https://github.com/wso2/dpdp-accelerator/commit/dc4f0c057eec6ae0ab532ccb5bda691b4d6016c5

docs/content/complaint-management-guide.md
```diff
@@ -0,0 +1,21 @@
+---
+title: Configuring Complaint Management
+sidebar_position: 4
+---
+## Customizing complaint notification emails
```
hasithakn: dont we already have a md guide for grievances? do you think we need separate doc file for this?

dpdp-accelerator/components/org.wso2.dpdp.accelerator.identity.extensions/pom.xml
```diff
@@ -237,8 +241,11 @@
                             org.apache.commons.logging;version="${commons.logging.imp.pkg.version.range}",
                             com.google.gson;version="${gson.imp.pkg.version.range}",
 
+                            org.apache.axiom.*;version="0",
```
hasithakn: org.apache.axiom.*;version="0"

version="0" is valid OSGi syntax, but it means "accept any axiom version at all" — no lower or upper bound — which defeats the point of pinning versions the way every other entry in this block does (;version="${some.version.range}").

It happens to match an existing line in common/pom.xml (line 154), so this isn't new. However, since we're touching Axiom import wiring here, it would be better to fix both while we're at it rather than adding another copy of the unpinned version.

Suggest adding a proper range property to the root pom.xml, next to axiom.version:

<axiom.version>1.2.11-wso2v16</axiom.version>
<axiom.imp.pkg.version.range>[1.2.11, 2.0.0)</axiom.imp.pkg.version.range>

Then use:

org.apache.axiom.*;version="${axiom.imp.pkg.version.range}"

in both identity.extensions/pom.xml (this PR) and common/pom.xml (pre-existing), matching the pattern already used for carbon.consent.mgt.core.imp.pkg.version.range.

docs/content/complaint-management-guide.md
ParameswaranSajeenthiran: Moved to the greivences-guide.md in https://github.com/wso2/dpdp-accelerator/pull/178/commits/ed5eac033d4404c57ef8a275758772f68de1007b

dpdp-accelerator/components/org.wso2.dpdp.accelerator.identity.extensions/pom.xml
ParameswaranSajeenthiran: Updated in https://github.com/wso2/dpdp-accelerator/pull/178/commits/ed5eac033d4404c57ef8a275758772f68de1007b

docs/content/localization-guide.md
```diff
@@ -1,6 +1,6 @@
 ---
 title: Localizing the Consent Portal
-sidebar_position: 4
+sidebar_position: 5
```
hasithakn: since no new pages this is not needed right? check below one as well

hasithakn (approved):

PR #192, opened 2026-09-11
Separated event API models from service DTOs

### LINKED ISSUE #39: Restore a dedicated REST DTO layer + OpenAPI spec for the Event Notification endpoints
## Summary

`TopicEndpoint`, `SubscriptionEndpoint`, and `EventEndpoint`
(`internal-webapps/org.wso2.dpdp.accelerator.event.notifications.endpoint`) currently take and
return `org.wso2.dpdp.accelerator.event.notifications.service.dto.*` classes directly as their
JAX-RS request/response types. There is no REST-layer DTO of its own — the service layer's internal
DTOs are doing double duty as the public wire format. A change to a service-layer DTO today silently
changes the public REST contract.

This module also has no OpenAPI spec, no codegen wiring, and no `dto`/`util` mapper package at all —
`git log` shows an `event-notifications.yaml` spec + generated `dto` package + a
`EventNotificationDtoMapper` did exist on an earlier branch, but never made it into the current
history; a later refactor rewrote the module without them.

The consent-history endpoint (`internal-webapps/org.wso2.dpdp.accelerator.consent.mgt.extensions.endpoint`)
already has the correct, working version of this pattern: an OpenAPI spec generates
request/response models only (`generateApis=false`, `jaxrs-cxf` generator) into the module's own
`dto` package under `src/gen/java`, kept separate from any service-layer types.

## Proposed change

Restore true separation for the event-notification endpoints, matching the consent-history pattern:

1. Author a fresh `event-notifications.yaml` OpenAPI spec from the *current* endpoint behavior
   (topics/subscriptions/events — the API surface has changed significantly since any earlier spec
   existed, so it can't be copied back as-is).
2. Wire up `openapi-generator-maven-plugin` + `build-helper-maven-plugin` in this module's `pom.xml`,
   same shape as the consent-history endpoint's.
3. Generate the endpoint-local `dto` package under `src/gen/java`.
4. Write a mapper (`util` package) converting service DTOs ↔ the new REST DTOs.
5. Update `TopicEndpoint`, `SubscriptionEndpoint`, and `EventEndpoint` to accept/return the new REST
   DTOs instead of service DTOs directly, calling the mapper at the boundary.
6. Update the existing tests for those three resource classes/handlers accordingly.

## Scope note

This is a bigger version of the consent-history work: 9 service DTOs, 3 resource classes, roughly a
dozen JAX-RS methods total, versus consent-history's 2 read endpoints. No DB, business-logic, or
config changes — REST-boundary only.

ThiwankaChanditha:
## Purpose

  Resolves a layering issue in the Event Notification REST module: public request and response models were service DTOs, so the external API contract was coupled directly to the service layer.

  ## Goals

  - Make event-notifications.yaml the source of truth for the public REST contract.
  - Generate endpoint-local API DTOs under endpoint.dto.
  - Keep service DTOs, handlers, business logic, DAO models, and persistence unchanged.
  - Map between generated API models and service DTOs only at the JAX-RS boundary.
  - Preserve existing JSON behavior, error responses, enum parsing, timestamp representation, secret suppression, and HMAC-signed raw request bodies.

  ## Approach

  The endpoint module now uses OpenAPI Generator (jaxrs-cxf, model-only generation) to generate 28 API models into src/gen/java. Generated sources are committed and compiled as
  endpoint-local models.

  EventNotificationDtoMapper converts:

  - Generated topic, subscription, and event create requests into service DTOs.
  - Service DTOs and paginated results into generated API response models.
  - Delivery history, attempts, polling responses, timestamps, and payload fields while preserving the prior wire representation.

  TopicEndpoint, SubscriptionEndpoint, and EventEndpoint now map requests before calling handlers and map service results before returning HTTP responses.

  Polling and delivery-completion requests continue to use their original raw request bodies. This preserves exact request bytes required for HMAC signature verification.

  Custom generated-enum deserializers preserve the existing case-insensitive, whitespace-tolerant parsing and aliases. The exception mapper now returns the generated Error API model.

  Legacy request metadata remains accepted for compatibility where existing handlers already ignore or derive those values from tenant context. Subscription response delivery
  settings continue to omit the shared secret.

  ## User Stories

  N/A — internal REST-boundary refactor. Existing API behavior and service contracts are preserved.

  ## Release Note

  Internal Event Notification REST API refactor:

  - Public DTOs are generated from the OpenAPI contract and owned by the endpoint module.
  - Service DTOs remain internal to the service layer.
  - No intended API behavior or wire-format changes.

  ## Documentation

  Added endpoint-module documentation covering:

  - OpenAPI regeneration.
  - Generated-source ownership.
  - Runtime mapper responsibilities.
  - Compatibility guarantees.
  - Packaged-WAR verification expectations.

  ## Training

  N/A.

  ## Certification

  N/A 

  ## Marketing

  N/A.

  ## Automation Tests

  Updated and added endpoint tests covering:

  - Request and response JSON compatibility.
  - Nested delivery history and pagination.
  - Epoch-millisecond timestamps and payload strings.
  - Shared-secret suppression.
  - Legacy request metadata.
  - Enum aliases and malformed requests.
  - Error response shape and codes.
  - Exact signed-body forwarding for polling and completion.

  Validation completed:

  - Endpoint Maven verification: 40 tests, 0 failures.
  - OpenAPI regeneration produced identical sources for 28 generated models.
  - Packaged WAR inspection confirmed generated DTOs, mapper, and deserializer classes.
  - git diff --check passed.
  - Live Identity Server deployment testing was not run.

  ## Security Checks

  - Shared secrets remain excluded from subscription response models and are covered by compatibility tests.
  - Polling and completion signatures continue to be verified against the original request body.
  - FindSecurityBugs and a formal secure-engineering review were not run.

This Fixes #39

### DIFF
--- a/dpdp-accelerator/internal-webapps/org.wso2.dpdp.accelerator.event.notifications.endpoint/src/main/openapi-templates/enumOuterClass.mustache
+++ b/dpdp-accelerator/internal-webapps/org.wso2.dpdp.accelerator.event.notifications.endpoint/src/main/openapi-templates/enumOuterClass.mustache
@@ -1,0 +1,51 @@
+{{#jackson}}
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonValue;
+{{/jackson}}
+
+/**
+ * {{description}}{{^description}}Gets or Sets {{{name}}}{{/description}}
+ */
+{{#vendorExtensions.x-class-extra-annotation}}
+{{{.}}}
+{{/vendorExtensions.x-class-extra-annotation}}
+{{>additionalEnumTypeAnnotations}}public enum {{{datatypeWithEnum}}}{{^datatypeWithEnum}}{{{classname}}}{{/datatypeWithEnum}} {
+  {{#gson}}
+  {{#allowableValues}}{{#enumVars}}
+  @SerializedName({{#isInteger}}"{{/isInteger}}{{#isDouble}}"{{/isDouble}}{{#isLong}}"{{/isLong}}{{#isFloat}}"{{/isFloat}}{{{value}}}{{#isInteger}}"{{/isInteger}}{{#isDouble}}"{{/isDouble}}{{#isLong}}"{{/isLong}}{{#isFloat}}"{{/isFloat}})
+  {{{name}}}({{{value}}}){{^-last}},
+  {{/-last}}{{#-last}};{{/-last}}{{/enumVars}}{{/allowableValues}}
+  {{/gson}}
+  {{^gson}}
+  {{#allowableValues}}{{#enumVars}}
+  {{{name}}}({{{value}}}){{^-last}},
+  {{/-last}}{{#-last}};{{/-last}}{{/enumVars}}{{/allowableValues}}
+  {{/gson}}
+
+  private {{{dataType}}} value;
+
+  {{{datatypeWithEnum}}}{{^datatypeWithEnum}}{{{classname}}}{{/datatypeWithEnum}}({{{dataType}}} value) {
+    this.value = value;
+  }
+
+  @Override
+{{#jackson}}
+  @JsonValue
+{{/jackson}}
+  public String toString() {
+    return String.valueOf(value);
+  }
+
+{{#jackson}}
+  @JsonCreator
+{{/jackson}}
+  public static {{{datatypeWithEnum}}}{{^datatypeWithEnum}}{{{classname}}}{{/datatypeWithEnum}} fromValue({{{dataType}}} value) {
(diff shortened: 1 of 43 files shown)


anjuchamantha (dismissed):

dpdp-accelerator/internal-webapps/org.wso2.dpdp.accelerator.event.notifications.endpoint/src/main/openapi-templates/enumOuterClass.mustache
anjuchamantha: Do we need this file?

anjuchamantha:
@coderabbitai review

hasithakn (approved):

Issue #204, opened 2026-09-15
[H2] All Consent page search filters not working properly


anjuchamantha:
Observed with H2 DB (Seems the issues are coming from IS)

- User search not working - all users' consents are shown
- Advance search date filter throws 400
- Advance search `Consent property` requires both key and value - after adding key, make value also mandatory

Check other search related features as well

PR #209, opened 2026-09-16
E2E suite: dual-profile foundation, coverage fixes, and CI speed-ups

### LINKED ISSUE #136: Document all test scenarios and steps, for assurance purposes
### What

Write down every test scenario we check (manual and automated) and the steps for each, in one
place.

### Why

Needed as a reference for quality assurance — so anyone can see exactly what's been tested and
how, without having to dig through test code.

anjuchamantha:
## Summary

Makes the E2E suite runnable against two Playwright profiles — a per-run throwaway tenant (`multi-tenant`) and the always-present super tenant (`super-tenant`) — with no pre-existing test needing to know which one it's running under, then uses that foundation to close real coverage gaps, fix real product bugs, and cut the suite's own CI runtime.

**Two-profile foundation**
- Adds `tenant-setup`/`user-setup`/`super-tenant-user-setup` Playwright projects and the SCIM2/console plumbing (`utils/serverConfig.ts`, `utils/tenantProvisioning.ts`, `utils/scimProvisioning.ts`, `utils/runState.ts`) each profile's setup needs.
- `super-tenant` no longer transitively depends on tenant creation — it used to create a throwaway tenant it never touched, just because its persona-provisioning test lived in the same file/project as the tenant-scoped one. Split into its own file (`01.03`) and project.
- Unifies persona usernames across both tenants (`dpdp-user-1@dpdp.test`, `dpdp-admin@dpdp.test`, etc.) — they only ever needed to be unique within their own tenant's user store, not globally distinct.

**Coverage gaps closed** (Elements, Purposes, Consents areas — properties/detail-view accuracy, pagination caps, deletion, purpose versioning, rejected-consent state, the Relation filter, and a new `03.05.05` proving a Purpose version referenced by a consent can't be deleted)

**Real product bugs fixed, each verified live in both directions (fails before the fix, passes after)**
- Doubled "Revoked by reviewer by `<actor>`" text in the self-service consent-history view.
- "Revoke Consents" (plural) in the revoke confirmation dialog, inconsistent with its own sibling actions ("Approve Consent", "Reject Consent").
- `AdminConsentFilters`'s draft-filter state was a single object shared between the always-visible toolbar fields (User, Consent ID) and the "Advanced filters" popover's own fields — opening or cancelling that popover silently discarded whatever had been typed into the toolbar fields but not yet applied. Related to #204 ("User search not working - all users' consents are shown").
- `VITE_AUTH_ENABLED` removed entirely: it was only ever set by the Maven build, never by a plain `npm run build` (no `.env`/`.env.production` committed), so a manual frontend build silently compiled out the entire OIDC login flow — no redirect to sign-in, every API call unauthenticated. Authentication is now unconditional.

**Root-caused, not just retried, an intermittent Consents test failure** — traced to two real, deterministic bugs (not flakiness): the Relation-filter query fell off its own default page once a shared persona's consent count passed one page (consents are never deleted in this suite), and `clearAllFilters()` didn't wait for its own filter-panel remount before the next fill, letting a just-typed value get silently wiped. Both fixed; verified with 8 consecutive clean runs.

**Local Playwright workers capped at 2** (`playwright.config.ts`) — the CPU-based default (4 on this 8-core dev machine) measurably caused resource-contention flakiness (session/auth timeouts, occasional browser-context crashes) running alongside WSO2 IS and MySQL on the same machine; repeated full-suite runs at 2 workers came back completely clean. CI's own runner is untouched.

**CI runtime cut roughly in half**
- Seeding for tests that aren't themselves exercising a create-form (pagination caps, act-on/search/history fixtures) now goes through the admin API instead of driving the real "Add Element"/"Add Purpose" UI forms every time — measured ~2x faster on the affected tests locally, and a real CI run went from 21m20s to 10m55s on the "Run Playwright" step alone after this change.
- `pr-e2e.yml` now runs only the `multi-tenant` project on every labelled PR — `super-tenant`'s own coverage (unqualified-root routing, tenant-provisioning skip) wasn't worth doubling every PR's E2E time for, so a new `nightly-e2e.yml` runs every project once a day instead. Release/weekly gates are unaffected (still run every project).
- `06-multi-tenancy` cut from 3 tests to 1: the other two either duplicated what a broken listener would already surface as mass downstream failures, or tested authorization mechanics `05-authorization` already covers generically. The remaining isolation test no longer creates its own throwaway tenant — it reuses the run's own per-run tenant and the super tenant directly.

**Suite organization**
- `04-consents` reordered so viewing → searching → acting reads the same for both personas (user, then admin) instead of "acting" sitting first for no reason.
- Merged/moved the two "happy path" element/purpose tests that lived oddly in the *viewing* files back into the *creating* files where every other area's happy path lives.
- `seedConsent` and four other API-only setup helpers renamed with a `ViaApi` suffix so the name says what they do, rather than relying on a docblock to explain it.
- `Elements`/`Purposes` are no longer deleted after tests (previously the only two record types that were) — they now accumulate like Consents/complaints already did, since nothing in this suite ever asserted on emptiness and the cleanup was pure unnecessary API traffic.

Fixes https://github.com/wso2/dpdp-accelerator/issues/136, https://github.com/wso2/dpdp-accelerator/issues/84

## Test plan
- [x] `npx tsc --noEmit` clean (test suite and frontend)
- [x] `npm run verify:ids` clean — 174 test IDs, 46 spec files, 9 areas
- [x] Frontend unit tests: 295/295 passing
- [x] Full E2E suite, both profiles, repeated runs: clean against a fresh WSO2 IS + MySQL pack
- [x] Rebased onto current upstream `main` (post docs-migration) — resolved conflicts by hand, re-verified after
- [ ] CI E2E run — needs a maintainer to apply the `Action/trigger-e2e` label per this repo's `pr-e2e.yml`/`pr-e2e-gate.yml` gating


dpdp-accelerator/react-apps/consent-portal/frontend/public/i18n/en/common.json
```diff
@@ -154,7 +154,7 @@
         "title": "Confirm Revocation",
         "message": "Are you sure you want to revoke consent?",
         "note": "This action revokes both mandatory and optional consents granted for all associated purposes.",
-        "confirm": "Revoke Consents"
+        "confirm": "Revoke Consent"
```
anjuchamantha, line 157: why this change is required?

dpdp-integration-test-suite/playwright.config.ts
```diff
@@ -17,30 +17,72 @@
...
-  // sessions - see that file for the full mechanism. No `workers` override is needed here as a
-  // result; Playwright's own CPU-based default applies.
+  // sessions - see that file for the full mechanism.
+  // Capped at 2 locally: each worker drives its own full Chromium instance alongside WSO2 IS and
+  // MySQL on this same machine, and the CPU-based default (half the detected cores, 4 on an 8-core
+  // Mac) was measurably causing resource-contention timeouts - a fixed, low worker count trades
+  // wall-clock time for a meaningfully lower flake rate. CI's own runner already computes to 1
+  // worker on its own (2 vCPUs / 2) via the untouched default - leave that alone rather than
+  // raising it to 2, since CI has even less headroom (2 vCPUs, 7 GB, shared with IS and MySQL) than
+  // this fix assumes.
```
anjuchamantha: No need these huge explanation.

dpdp-accelerator/react-apps/consent-portal/frontend/public/i18n/en/common.json
anjuchamantha: Real bug, not an arbitrary edit — `revocation.confirm` said "Revoke Consents" (plural) while its two siblings, `approval.confirm` ("Approve Consent") and `rejection.confirm` ("Reject Consent"), are both singular. The shared ConsentActionDialog only ever confirms one consent at a time, so the plural was simply wrong text.

Found it because ConsentDetailPage's locator for this button used non-exact matching, so the page object's own correct constant ("Revoke Consent") was silently substring-matching the wrong deployed text ("Revoke Consents"). Tightened the locator to `exact: true` while adding new Consents coverage, which then correctly failed against the still-plural text — verified live in both directions (fails before this fix, passes after) on both profiles. Commit bb3a0c0 has the full trace.

dpdp-integration-test-suite/playwright.config.ts
anjuchamantha: Fair, trimmed it down to 2 lines. Also caught that it had a stale/wrong claim while I was in there — CI's ubuntu-latest runner is actually 4 vCPUs, not 2, so it already computes 2 workers by default via Playwright's own half-the-cores heuristic (verified from a real CI run log), not 1 as the old comment assumed. Fixed in 77f46f5.

anjuchamantha:
## Why the E2E check is failing

Every single test failed identically with:
```
Error: "personas.dpo.password" is not configured. Set it in e2e-config.local.json (or e2e-config.json), or run ./scripts/setup-local.sh to generate it.
```
This is a suite-wide setup failure, not a product/test bug — nothing gets far enough to actually exercise the accelerator.

**Root cause:** `e2e.yml`'s own inline password-generation step only generates passwords for `('user', 'user2', 'consentAdmin')`, never `dpo`, so the `dpo` persona this branch introduced has no password by the time Playwright starts. This PR *does* fix that (the loop now includes `'dpo'`) — but `pr-e2e.yml` triggers via `pull_request_target`, which GitHub always resolves against the **base** repository's copy of the called workflow (`e2e.yml` on `wso2/dpdp-accelerator:main`), never the PR branch's own version. That's a deliberate GitHub security boundary (a PR can't use `pull_request_target` to run its own modified workflow with the base repo's secrets), not something fixable from within this PR.

Confirmed: `wso2/dpdp-accelerator:main`'s current `e2e.yml` still has the old, unfixed loop. This PR's own E2E check therefore can't self-validate its own workflow fix — it will keep failing this exact way on every re-run of the label until the PR merges and `e2e.yml`'s fix lands on `main` itself.

**Already validated independently**, since the PR's own check couldn't: pushed this branch's `e2e.yml` fix directly to the fork's own `main` (anjuchamantha/dpdp-accelerator) and ran the workflow there, where it *does* use the fixed file — full suite passed clean on both profiles.

hasithakn (approved):

Issue #243, opened 2026-09-18
Refactor sql exceptions throwing logic


hasithakn:
## Summary

Audited exception handling across the three main service boundaries — `event.notifications`, `complaint.mgt`, and `consent.mgt.extensions` — against the standard Java "Exception Translation" / "Fault Barrier" pattern (Effective Java Item 73; Core J2EE Patterns). The current design has the right *shape* but two real gaps: no classification happens in the middle of the chain, and the "don't leak DB details to clients" guarantee is enforced by convention only, not by the compiler or CI.

## What's already correct

- **Exception Translation at the DAO layer**: every DAO catches `SQLException` and throws a module-specific unchecked exception (`EventNotificationDataAccessException`, `ComplaintDAOException`, `ConsentHistoryDataInsertionException`/`RetrievalException`), keeping the original as `cause`. This is the textbook pattern, done correctly.
- **A Fault Barrier exists**: each webapp has exactly one `ExceptionMapper` (`EventNotificationExceptionMapper`, `ComplaintExceptionMapper`, `ConsentHistoryEndpointExceptionMapper`) that every exception must pass through before becoming an HTTP response.
- **Verified, today, no leak**: audited every catch site across all three modules that bridges a DAO-level exception to a "trusted" domain exception (`EventNotificationException`/`ComplaintException`/`ConsentHistoryEndpointException`). Every single one constructs the domain exception with a hardcoded, developer-written message; the real `SQLException`/DAO exception is only ever attached as `cause` (invisible to the client, visible only in server logs) or passed into `LogSanitizer`-wrapped log lines. No constructor of these three exception types auto-derives its message from a wrapped cause. Confirmed clean across ~40+ relevant catch blocks.

## Gaps

1. **No translation happens in the middle of the chain.** A DAO-level exception (e.g. `EventNotificationDataAccessException`) is never caught and re-classified by the service layer — it rides unchanged through service → handler → API resource, and only gets "handled" once it reaches the final `ExceptionMapper`'s generic catch-all. Because nothing classifies it along the way, every kind of unexpected failure (DB outage, bad SQL, connection pool exhaustion, a genuine bug) collapses into the exact same undifferentiated `INTERNAL_SERVER_ERROR`, with no way for logs/monitoring to tell them apart.
[...]

PR #264, opened 2026-09-21
Refactor exception handling: shared self-describing base, consistent DAO/service naming

hasithakn:
## Summary

Refs #243

- Introduces `DPDPException` (`common` module) as the shared self-describing base (`errorCode`/`description`/`httpStatus`) for every domain exception across `complaint.mgt`, `event.notifications`, and `consent.extensions`, set at the exact point each exception is thrown rather than attached later by whichever catch translates it.
- Introduces `DPDPSystemException` for system-level failures that never carry a response shape — `dpdp-accelerator.xml` parsing, OSGi bundle-activation DB connectivity checks, JDBC commit/connection failures, and a missing Carbon tenant context. None of these have a module-specific `ExceptionMapper` reading a code/status off them, so they stay plain.
- Standardizes naming and package layout per module: a `<Module>DAOException`/`<Module>DaoException` at the DAO layer, a `<Module>ServiceException` at the service layer, each rooted in `DPDPException` within its own module.
- `consent.extensions`: converted its three DAO exceptions from checked to unchecked, and added a new `ConsentExtensionsServiceException` so every service method now catches its own module's DAO exception and translates it before it can leave the service layer — previously these propagated raw, all the way to the endpoint's generic fallback. This closes the "no translation happens in the middle of the chain" gap from #243 for this module.
- Relocated error-code/constants classes that had drifted into `exception` packages (`ComplaintErrorCode`, `ComplaintServiceConstants`) into proper `constants` packages.
- Simplified `consent.mgt.extensions.endpoint`'s response building: removed a redundant endpoint-level re-wrap of an already-correctly-shaped service exception, widened its `ExceptionMapper` to read that exception directly instead — verified this produces byte-identical API responses to before.
- Updated `CLAUDE.md`'s exception-handling conventions to match.

## What this PR does not cover

Per #243's own scope, three items are intentionally left for follow-up:
- The same DAO→service translation gap identified in this PR for `consent.extensions` also exists in `complaint.mgt` and `event.notifications` (audited, not yet fixed).
- Hardened exception-constructor factory methods (e.g. `ComplaintException.wrapping(errorCode, cause)`) that remove the raw-message parameter.
- An automated guardrail (ArchUnit/PMD/SpotBugs) enforcing that a domain exception's message is never derived from a caught exception's own message.

## Test plan

- [x] `mvn clean install` from the repository root — full reactor, all modules, `BUILD SUCCESS`
- [x] Verified no API response (error code, HTTP status, or message) changed for any existing, exercised code path
- [x] Added/updated unit tests covering the new DAO→service translation behavior in `consent.extensions`

🤖 Generated with [Claude Code](https://claude.com/claude-code)


dpdp-accelerator/components/org.wso2.dpdp.accelerator.common/src/main/java/org/wso2/dpdp/accelerator/common/persistence/DBDialectConstants.java
```diff
@@ -0,0 +1,38 @@
...
+ * isn't redefined per module. Only the dialects an accelerator dbscript actually ships for today
+ * are listed - add another constant here only once a feature ships a dedicated
+ * {@code dbscripts/<feature>/<dialect>.sql} for it.
+ */
+public final class DBDialectConstants {
+
+    public static final String DB_TYPE_MYSQL = "mysql";
+    public static final String DB_TYPE_POSTGRES = "postgres";
+    public static final String DB_TYPE_SQLITE = "sqlite";
+    public static final String DB_TYPE_DEFAULT = "default";
```
anjuchamantha, line 34: Do we need a default? is it H2?

anjuchamantha (dismissed):

dpdp-accelerator/components/org.wso2.dpdp.accelerator.common/src/main/java/org/wso2/dpdp/accelerator/common/persistence/DBDialectConstants.java
hasithakn: Good catch — `"default"` was never a real value, it was a fallback key for "we don't know the dbType," and it happened to route to the same class H2 also fell into by coincidence, which made it easy to misread as "H2 needs this default."

Fixed it properly instead of just clarifying in a comment: added a dedicated `DB_TYPE_H2` constant and an explicit `*H2DBQueries` provider per module (extending the common/baseline class, same empty-override-point pattern already used for `*MysqlDBQueries`), and removed `DB_TYPE_DEFAULT` entirely. The "unknown dbType" fallback now resolves to H2 explicitly, since H2 is genuinely the accelerator's own embedded default (`configure.sh`'s `DB_TYPE=h2`), rather than a made-up sentinel string. H2 is now named and routed the same way mysql/postgres/sqlite already are, with its own override point reserved for a future H2-specific quirk. Also moved consent-expiry's oracle/sql server/mssql literals into `DBDialectConstants` for consistency.

Pushed in d191f6d8 across all 4 query factories (`ConsentExpiryQueryFactory`, `ConsentHistoryQueryFactory`, `ComplaintQueryFactory`, `EventNotificationQueryFactory`) plus test coverage for the new branch. Full reactor build passes.

anjuchamantha (approved):

